PR #2, opened 2024-08-15
story/VOGRE-2

Girik1105:
[VOGRE-2](https://diging.atlassian.net/browse/VOGRE-2?atlOrigin=eyJpIjoiOWQ5OTZkMzRmNGJkNDYyNThhOTJjZjNmOGQ4YWM4ZTMiLCJwIjoiaiJ9)

Changed made:
- Changed from GitHub Auth to Local Authentication
- Fixed Spacing in settings.py
- Fixed Local Environment Variable Config
- Made local auth templates pretty

[VOGRE-2]: https://diging.atlassian.net/browse/VOGRE-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jdamerow:
that file should not have been pushed, so I deleted it and added it to `.gitignore` on develop. Please delete the file, that should make the conflict go away.

jdamerow:
There seems to be a dependency missing:
`ModuleNotFoundError: No module named 'django_inlinecss'`

Girik1105:
I added the collectstatic configuration because for some reason django_inlinecss only references STATIC_ROOT for locating CSS files and not STATICFILES_DIRS(which is used in development).

jdamerow:
- email should be required when registering
- when trying to register I get:
```
'NoneType' object has no attribute 'rsplit'
Request Method: | POST
-- | --
http://localhost:8000/signup/
2.2.20
AttributeError
'NoneType' object has no attribute 'rsplit'
/usr/local/lib/python3.7/site-packages/django/utils/module_loading.py in import_string, line 13
```
It looks like it's when it tries to send an email.
- and the signup/log in pages layout needs to match the overall one. However, we can wait with this until the css of the home page is fixed and we probably do some more adjustments.

Girik1105:
> * email should be required when registering
> * when trying to register I get:
> 
> ```
> 'NoneType' object has no attribute 'rsplit'
> Request Method: | POST
> -- | --
> http://localhost:8000/signup/
> 2.2.20
> AttributeError
> 'NoneType' object has no attribute 'rsplit'
> /usr/local/lib/python3.7/site-packages/django/utils/module_loading.py in import_string, line 13
> ```
> 
> It looks like it's when it tries to send an email.
> 
> * and the signup/log in pages layout needs to match the overall one. However, we can wait with this until the css of the home page is fixed and we probably do some more adjustments.

I think this error is arising because Django-allauth is trying to send an email but does not have the right credentials, please add the following credentials in your .env file or directly in settings.py

to check the email in your terminal:
EMAIL_BACKEND=django.core.mail.backends.console.EmailBackend

to get actual emails from the app:
EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
EMAIL_HOST = 'smtp.gmail.com' (or whichever service you're using)
EMAIL_PORT = 587
EMAIL_USE_TLS = True
EMAIL_HOST_USER = # Your Gmail address
EMAIL_HOST_PASSWORD = #gmail app password, gmail only allows app passwords for sending emails through dev envs (https://support.google.com/accounts/answer/185833?hl=en)

PR #6, opened 2024-09-27
story/VOGRE-3

Girik1105:
Changes:

Refactored Repository Model, manager, auth and views to handle citesphere text imports
Removed excess repository code
Fixed Annotation files (vue.html, resources.js, text.js) to create annotations, appellations, relations

Before Testing:
Please configure a repository and connect your account to citesphere, as we changed the repository to only use citesphere, no additional configuration would be required.


annotations/annotators.py
```diff
@@ -100,15 +100,25 @@ def get_content(self, resource):
...
+
+        if not self.text:
+            return None
+
+        # Directly retrieve the tokenized content from the Text model.
+        try:
+            # Retrieve the tokenized content of the text
+            resource = self.text.tokenizedContent
+            self.resource = resource
+            return self.resource
```
jdamerow: are these two lines necessary? why not `return self.text.tokenizedContent`? or at least `self.resource = self.text.tokenizedContent`?

annotations/annotators.py
```diff
@@ -100,15 +100,25 @@ def get_content(self, resource):
...
+        except Text.DoesNotExist:
```
jdamerow: where is that being thrown? Isn't that thrown from a db transaction? Is there a db call made here?

annotations/annotators.py
```diff
@@ -152,22 +162,44 @@ def render_display(self, context={}):
...
-        detect  = chardet.detect(content)
-        return {
+        
+        # Extract the word_id from the request
+        word_id = request.GET.get('word_id')
+        
+        # Find the word with the specified ID in the tokenized content
+        if word_id:
+            # Using regex to find the specific word with the given word_id
+            import re
```
jdamerow: why is that not imported at the beginning of the file?

annotations/annotators.py
```diff
@@ -152,22 +162,44 @@ def render_display(self, context={}):
...
+            word_pattern = re.compile(r'<word id="' + re.escape(word_id) + r'">(.*?)<\/word>')
+            match = word_pattern.search(resource)
+            if match:
+                specific_word = match.group(1)  # Extract the word text
+            else:
+                specific_word = None
+        else:
+            specific_word = None
```
jdamerow: just define this before `if word_id` and then you don't need the two else statements above

annotations/models.py
```diff
@@ -1147,3 +1150,16 @@ class DocumentPosition(models.Model):
     If :attr:`.position_type` is :attr:`.WHOLE_DOCUMENT`\, then this can be
     blank.
     """
+
+
+class ImportedCitesphereItem(models.Model):
```
jdamerow: what is this class for?

annotations/static/annotations/js/annotators/resources.js
```diff
@@ -2,7 +2,41 @@
...
+      for (let i = 0; i < cookies.length; i++) {
+          const cookie = cookies[i].trim();
+          if (cookie.substring(0, name.length + 1) === (name + '=')) {
+              cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+              break;
+          }
+      }
+  }
+  return cookieValue;
+}
```
jdamerow: this looks like duplicate code

annotations/static/annotations/js/annotators/resources.js
```diff
@@ -2,7 +2,41 @@
...
+
+Vue.http.headers.common['X-CSRFTOKEN'] = csrftoken;
+
+var BASE_URL = 'http://localhost:8000';
```
jdamerow: that should not be hard-coded

annotations/static/annotations/js/annotators/text.js
```diff
@@ -346,12 +346,14 @@ ConceptPickerItem = {
...
                 <div>
-                    <a v-on:click="select" style="cursor: pointer;">{{ concept.interpretation_label }} ({{ concept.interpretation.authority }})</a>
+                    <a v-on:click="select" style="cursor: pointer;">{{ concept.interpretation.label }}</a>
+                    <p>{{ concept.interpretation.uri }}</p>
                 </div>
                 <div class="text text-muted">{{ concept.interpretation.description }}</div>
             </div>`,
     methods: {
         select: function () {
+            console.log(this.concept)
```
jdamerow: remove

jdamerow (changes requested):
I don't seem to be able to import texts, and I think there is something off with the model etc. we might need to talk about this.

.env-example
```diff
@@ -9,4 +9,8 @@ EMAIL_USE_TLS=
 CITESPHERE_ENDPOINT=
 CITESPHERE_CLIENT_ID=
 CITESPHERE_CLIENT_SECRET=
```
jdamerow: those three above  should not be needed anymore

annotations/models.py
jdamerow: still confused if this is needed

annotations/views/rest_views.py
```diff
@@ -483,17 +492,67 @@ def search(self, request, **kwargs):
...
-            _fields = {
-                'name': 'label',
-                'id': 'alt_id',
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in list(datum.items())}
-        return Response({'results': list(map(_relabel, [c.data for c in concepts]))})
+        url = f"{settings.CONCEPTPOWER_ENDPOINT}ConceptLookup/{q}/{pos if pos else ''}"
+        response = requests.get(url, auth=(settings.CONCEPTPOWER_USERID, settings.CONCEPTPOWER_PASSWORD))
```
jdamerow, line 496: conceptpower returns json as well, so that should be used, but I created another ticket for that

repository/managers.py
```diff
@@ -1,17 +1,112 @@
...
         headers = {}
         if self.user:
-            headers = auth.jars_github_auth(self.user)
+            headers = auth.citesphere_auth(self.user)
         return requests.get(target, headers=headers, params=params).content
+
+    def groups(self):
+        """Fetch Groups from the repository's endpoint"""
+        headers = auth.citesphere_auth(self.user)
+        url = f"{self.repository.endpoint}/api/v1/groups/"  # Use repository's endpoint
```
jdamerow: you probably want the path in the settings file. or at least in a different class specifically for v1 of the api. If/when we move to a version 2 of the api, we want this to be fairly easily exchangable. thinking about this though, I'll create a new ticket for that!

annotations/views/citesphere_import_views.py
```diff
@@ -0,0 +1,25 @@
+from django.shortcuts import render, get_object_or_404, redirect
+from django.http import HttpResponseBadRequest
+from external_accounts.models import CitesphereItem
+from ..models import ImportedCitesphereItem, TextCollection
+from django.contrib.auth.decorators import login_required
+
+@login_required
+def import_citesphere_items_from_group(request, citesphere_item_id, project_id):
```
jdamerow: is this needed? for what?

jdamerow (changes requested):

jdamerow:
There are a couple of things for which I created new tickets. For this one to be merged, besides the few comments I made, the main thing it the connection to citesphere. Right now, I think when I browse a repository to annotate a text it only shows me what i have looked at before in the Dashboard? Is that correct? It should on loading the page, go out to citesphere and get a current list of groups, and when clicking on a group, get a current list of collections and items.

jdamerow:
And actually the same for the dashboard (load from citesphere on page load). We'll worry about caching later.

Girik1105:
> There are a couple of things for which I created new tickets. For this one to be merged, besides the few comments I made, the main thing it the connection to citesphere. Right now, I think when I browse a repository to annotate a text it only shows me what i have looked at before in the Dashboard? Is that correct? It should on loading the page, go out to citesphere and get a current list of groups, and when clicking on a group, get a current list of collections and items.

Citesphere data is rendered in two main places:

### 1. **Dashboard**
This is where users can **browse** through their Citesphere **Groups**, **Collections**, and **Items**. Users access this section right after connecting to Citesphere. When they browse, the groups, collections, and items are stored in the following models:

- **`CitesphereGroup`**
- **`CitesphereCollection`**
- **`CitesphereItem`**

These models are primarily used in conjunction with the following views in the **External Accounts app**:

- **`get_citesphere_groups(request)`**
- **`get_citesphere_collections(request, group_id)`**
- **`list_citesphere_groups(request)`**
- **`group_detail(request, slug)`**

This section mainly allows users to explore their Citesphere data, which is saved into the models when they browse.

### 2. **Repository Manager (Repository App)**
The **Repository Manager** is where direct API calls to Citesphere are made. This loads the **latest data** when the page is opened and displays it directly from the API responses. When users choose to **import a text** from Citesphere, the data is stored into the **`Text`** model for further annotation.

There is **no direct interaction** between the Repository Manager and the **CitesphereGroup**, **CitesphereCollection**, or **CitesphereItem** models from the **External Accounts app**. The only common link between the two is the **`CitesphereAccount`** model used for authentication, along with the associated authentication views and middleware.

.env-example
Girik1105: *`CITESPHERE_ENDPOINT`** is still being used in the **dashboard** for making API calls to Citesphere, specifically in views such as:

- **`get_citesphere_groups(request)`**
- **`get_citesphere_collections(request, group_id)`**

Additionally, **`CITESPHERE_CLIENT_ID`** and **`CITESPHERE_CLIENT_SECRET`** are still required for the **OAuth login flow** in the following views:

- **`citesphere_login(request)`**
- **`citesphere_callback(request)`**

jdamerow: no, these should all be configured when adding a repository through admin. there should be no general citesphere instance configured.

jdamerow (changes requested):
- CitesphereGroup
- CitesphereCollection
- CitesphereItem
should be deleted. They just blow up our database and model, and the user really doesn't need to browse Citesphere after connecting it (and if we want that, it doesn't need to cache at this point).

jdamerow:
There are still issues, but I'll merge this for now and will create new tickets, so Rajvi can also start working on this.

PR #7, opened 2024-10-10
bug/VOGRE-10

Girik1105:

- [ ] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed

Texts are not shown for Groups

Description
When importing from Citesphere, texts that are not in a collection but directly in a group also need to be shown.

https://diging.atlassian.net/browse/VOGRE-10?atlOrigin=eyJpIjoiYzQ3NDAxMWUzODRhNDFmYzlhZTYyNzA1YTYwZGYwNDgiLCJwIjoiaiJ9

This ticket dosen't directly depend on another ticket, however to maintain consistency I changed the `group_items` method in the `RepositoryManager` to show the all the texts in the group (more than 50) which is been done in `bug/VOGRE-13`
 
 https://diging.atlassian.net/browse/VOGRE-13?atlOrigin=eyJpIjoiMmZlMTI0NDZlMjU4NGIxZThlYzk0OWExNmNjYTkzZTQiLCJwIjoiaiJ9
  
- Added the `group_items` method in the `RepositoryManager` to fetch items directly within a group.
- Added the `repository_group_texts` view to use the `group_items` method to retrieve texts for a specific group.
- Implemented pagination in the `repository_group_texts` view, displaying 20 texts per page, and handled edge cases for page navigation such as invalid number pages.
- Added similar pagination logic for displaying texts in repository_collection_texts view to maintain consistency.


annotations/views/repository_views.py
```diff
@@ -239,18 +240,67 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
         return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
 
     project_id = request.GET.get('project_id', None)
+    
+    # Pagination
+    page = request.GET.get('page', 1)
+    paginator = Paginator(texts['items'], 20)  # Show 20 texts per page
```
jdamerow: 20 should be a configuration property via settings

annotations/views/repository_views.py
```diff
@@ -239,18 +240,67 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
...
+@citesphere_authenticated
+def repository_group_texts(request, repository_id, group_id):
+    """View to fetch and display texts within a group from Citesphere"""
+    user = request.user
+    repository = get_object_or_404(Repository, pk=repository_id)
+    manager = RepositoryManager(user=user, repository=repository)
+
+    try:
+        texts = manager.group_items(group_id)
+    except Exception as e:
```
jdamerow: I assume this also catches cases like "there is no such group" or "no permission"? In that case a 404 should be returned (or a forbidden).

annotations/views/repository_views.py
```diff
@@ -239,18 +240,67 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
...
+        return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
+
+    project_id = request.GET.get('project_id', None)
+    
+    # Pagination
+    page = request.GET.get('page', 1)
+    paginator = Paginator(texts['items'], 20)  # Show 20 texts per page
```
jdamerow: same as above, should be a configuration

repository/managers.py
```diff
@@ -30,6 +30,62 @@ def groups(self):
...
+        group_response = requests.get(base_url, headers=headers)
+        if group_response.status_code != 200:
+            group_response.raise_for_status()
+
+        # Parse the response to get the number of items in the group
+        group_data = group_response.json().get('group', {})
+        group_num_items = group_data.get('numItems', 0)
+
+        # Get total pages required to fetch all items (50 items per page)
+        total_pages = (group_num_items // 50) + (1 if group_num_items % 50 else 0)
```
jdamerow: same as before, should be configurable

repository/managers.py
```diff
@@ -30,6 +30,62 @@ def groups(self):
         else:
             response.raise_for_status()
 
+    def group_items(self, groupId):
```
jdamerow: do I understand this correctly, that this method returns **all** texts in a group? and then the view paginates? we have groups with multiple thousand texts in it, so this is not the right approach. Vogon should only get the page it wants to show from Citesphere.

jdamerow (changes requested):

annotations/views/repository_views.py
Girik1105: Currently in the manager function this is how an error is being handled:
```
        if response.status_code == 200:
            return response.json()  # Return the groups data
        else:
            response.raise_for_status()
```
If an API response is not 200, it will return a generic HTTPError. Then the subsequent view shows an error page. This is set for all manager functions and its following views. To maintain consistency, can I edit the error handling in [VOGRE-9](https://diging.atlassian.net/browse/VOGRE-9?atlOrigin=eyJpIjoiMDM3MTM0YWE4ZDI1NDQxY2I0NDc2Mjk2ZjE3OGYxMjUiLCJwIjoiaiJ9) since im editing all manager functions, I can change it handle specific errors and show different error messages and apply it all of the functions and views.

jdamerow: ya, that's fine

annotations/views/repository_views.py
```diff
@@ -239,18 +241,83 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
...
+    # range(total_items) is being used as a placeholder to create correct pagination metadata
+    paginator = Paginator(range(total_items), items_per_page)
+    try:
+        paginated_texts = paginator.page(page)
+    except PageNotAnInteger:
+        page = 1
+        paginated_texts = paginator.page(page)
+    except EmptyPage:
+        page = paginator.num_pages
+        paginated_texts = paginator.page(page)
```
jdamerow: I'm confused, why is this needed? If the user asks for a page that doesn't exist or is not a number, citesphere should either throw an error or return an empty list, so why would you need to check this here again? (in other words, I don't think you need a paginator object?)

jdamerow (changes requested):

annotations/utils.py
```diff
@@ -34,3 +34,36 @@ def basepath(request):
...
+        total_items (int): Total number of items.
+        page (int): The current page number requested.
+        items_per_page (int): Number of items per page.
+    
+    Returns:
+        dict: Pagination metadata including total pages, page range, and the validated current page.
+    """
+
+    # Calculate total pages by dividing total items by items per page. If there are any leftover items, an additional page is created.
+    total_pages = (total_items + items_per_page - 1) // items_per_page
```
jdamerow: I don't understand that, why does this take the total number of items and adds another page to ?

Girik1105: The problem is if there are leftover items that don’t perfectly fit into the available pages.

100 items, 20 items per page: 100 // 20 = 5
101 items, 20 items per page: 100// 20 = 5 (There should be 6 here but we miss one due to floor division)

`(total_items + items_per_page - 1) // items_per_page` solves this issue by forcing the integer division to round up whenever there are leftover items.

(101+20−1)//20=120÷20=6 pages

jdamerow: why not just do `math.ceil(total_items/items_per_page)`? with a comment about rounding up to accommodate "incomplete"  last pages. a lot easier to understand.

Girik1105: it totally slipped my mind that math.ceil does exactly that 🤦‍♂️,  I'll make the changes.

jdamerow (changes requested):
I still don't see any texts when I go to a group, just collections. I have to click on a collection to see texts.

annotations/views/repository_views.py
```diff
@@ -87,26 +88,29 @@ def repository_collections(request, repository_id):
...
         group_info = response_data.get('group')
         collections = response_data.get('collections', [])
+        group_texts =  manager.group_items(groupId=group_id, page=page)
     except IOError:
         return render(request, 'annotations/repository_ioerror.html', {}, status=500)
 
     project_id = request.GET.get('project_id')
     
-    base_url = reverse('repository_collection', args=(repository_id, group_id))
+    items_per_page = getattr(settings, 'PAGINATION_PAGE_SIZE')
```
jdamerow: Why don't you just use `settings. PAGINATION_PAGE_SIZE` and use it directly in line 113?

jdamerow (changes requested):

annotations/views/repository_views.py
Girik1105: I changed it to `settings.PAGINATION_PAGE_SIZE`

jdamerow:
Besides that should be good

PR #8, opened 2024-10-11
story/VOGRE-12

Girik1105:

- [ ] I have added Javadoc/documentation where appropriate

This ticket requires to show the user all files associated with a specific text in citesphere and allow the user to import any inidivudal file the user would like.

# Multiple texts for a Citesphere entry

Description:
Citesphere entries can have more than one text attached to them (right now I think the first one will be used?).
The text table should list all the file names that are attached to a record, and then instead of an “Import”  button, there should be an import icon next to each file.

https://diging.atlassian.net/browse/VOGRE-12?atlOrigin=eyJpIjoiNWJjMWNmMGJiYWYyNDdmMzkyYmIyZWMyMGZiMTA0MTEiLCJwIjoiaiJ9

No
  
I Updated the `repository_text_import` view to support file-specific imports by using `file_id` as part of the unique identifier (`uri`). This change ensures that each file import, even for the same `text_key`, creates or updates a distinct `Text` object. The view now returns `JsonResponse`, providing error messages and a `redirect_url` for successful imports, since now all files are being rendered via AJAX.

### DIFF
--- a/repository/managers.py
+++ b/repository/managers.py
@@ -52,7 +52,46 @@ class RepositoryManager(RESTManager):
         else:
             response.raise_for_status()
 
-    def item(self, groupId, itemId):
+    def item_files(self, groupId, itemId):
+        """
+        Fetch individual item from repository's endpoint and list all associated files for import.
+
+        Args:
+            groupId: The group ID in the repository
+            itemId: The item ID in the repository
+
+        Returns:
+            A dictionary containing a list of files with their respective details.
+        """
+        headers = auth.citesphere_auth(self.user, self.repository)
+        url = f"{self.repository.endpoint}/api/v1/groups/{groupId}/items/{itemId}/"
+        response = requests.get(url, headers=headers)
+
+        if response.status_code == 200:
+            item_data = response.json()
+
+            files = []
+            
+            # Extract Giles upload file details if available
+            giles_uploads = item_data.get('item', {}).get('gilesUploads', [])
+
+            if giles_uploads:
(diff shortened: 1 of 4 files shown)


repository/managers.py
```diff
@@ -69,6 +108,7 @@ def item(self, groupId, itemId):
 
         if response.status_code == 200:
             item_data = response.json()
+            print(item_data)
```
jdamerow, line 111: remove

jdamerow (changes requested):
Not sure, where this is coming from but I get this error:
`Failed to retrieve Giles document details: Invalid URL 'Noneapi/v2/resources/files/FILEnuoF0dESGftl/content/': No schema supplied. Perhaps you meant http://Noneapi/v2/resources/files/FILEnuoF0dESGftl/content/?`
(when I try to load a text)

Girik1105:
> Not sure, where this is coming from but I get this error: `Failed to retrieve Giles document details: Invalid URL 'Noneapi/v2/resources/files/FILEnuoF0dESGftl/content/': No schema supplied. Perhaps you meant http://Noneapi/v2/resources/files/FILEnuoF0dESGftl/content/?` (when I try to load a text)

I missed adding the `GILES_ENDPOINT` in the env-example. I just updated the example file. Once you add the giles geco staging endpoint in env it should resolve the issue!

jdamerow:
There don't seem to be any commit addressing the PR comments.

Girik1105:
My latest commits are not showing up on this PR, I'll just create a new PR.

PR #11, opened 2024-10-17
story/VOGRE-14

Girik1105:

Projects and texts

Description
The text import button should be for projects, not in the menu right now. Or if it stays in the main menu, the user should have to select a project to put the text in.
Actually, I think the way it currently is something is off. Texts should be part of projects, so the user goes to a project and sees all texts in that project. Not under “Texts”.

[VOGRE-14](https://diging.atlassian.net/browse/VOGRE-14?atlOrigin=eyJpIjoiZDM2M2VjODgwNTA0NGY3ZWI4YjA4NWIxYWJlYTgwNjgiLCJwIjoiaiJ9)

------
Projects should be sharable with other people.

Description
I want to be able to another user to my project and then they have access to all the texts and annotations in the project. And they should be able to add new texts and annotate them.

[VOGRE-15](https://diging.atlassian.net/browse/VOGRE-15?atlOrigin=eyJpIjoiNzE5MGY5NTc1YjIzNGQ3ZmIwYjczMmM1NzAwM2Q2N2YiLCJwIjoiaiJ9)

---

users should only see their own projects and projects that have been shared with them

[VOGRE-24](https://diging.atlassian.net/browse/VOGRE-24?atlOrigin=eyJpIjoiOTVjM2QwYjg2ZmE0NGQ2ODkzNDc0NDU2ZDkxZDZmMTciLCJwIjoiaiJ9)

- Text URLS now contain the project instead of being passed as parameters.
- Since there shouldn't be an instance where a text exists without a project, a user will be prompted to select a project when he is importing a text directly from a repository. A user can also directly add texts from a project.
- Collaborators will be able to view existing texts and the annotations on them when added to a project as a collaborator, and if a collaborator is removed they will no longer has access to the text but their annotations won't be deleted

[VOGRE-14]: https://diging.atlassian.net/browse/VOGRE-14?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annotations/views/project_views.py
```diff
@@ -36,6 +36,7 @@ def view_project(request, project_id):
     order_by = request.GET.get('order_by', 'title')
     texts = project.texts.all().order_by(order_by)\
                          .values('id', 'title', 'added', 'repository_source_id')
+    print(texts)
```
jdamerow: remove

annotations/templates/annotations/project_list.html
```diff
@@ -25,8 +25,13 @@
 
     <div class="list-group">
         {% for project in projects %}
-        <a class="list-group-item"
-            href="{% url "view_project" project.id %}">
+            {% if next_url %}
+                <a class="list-group-item"
+                href="{{ next_url }}?project_id={{ project.id }}">
```
jdamerow: why does the `project.id` need to be passed here?

annotations/views/repository_views.py
```diff
@@ -209,23 +209,34 @@ def repository_list(request):
 @citesphere_authenticated
 def repository_details(request, repository_id):
     template = "annotations/repository_details.html"
-    user = None if isinstance(request.user, AnonymousUser) else request.user
     repository = get_object_or_404(Repository, pk=repository_id)
+    user = None if isinstance(request.user, AnonymousUser) else request.user
+    project_id = request.GET.get('project_id')
```
jdamerow: merge with line 234

annotations/views/repository_views.py
```diff
@@ -209,23 +209,34 @@ def repository_list(request):
...
+
+    # Fetch text collections (projects) owned by the current user
+    user_owned_collections = TextCollection.objects.filter(ownedBy=user)
+    texts_by_project = {}
+
+    for collection in user_owned_collections:
+        # Filter texts within each collection that belong to the repository
+        collection_texts = collection.texts.filter(repository=repository).order_by('-added')
+        if collection_texts.exists():
+            texts_by_project[collection] = list(collection_texts)
```
jdamerow: does this mean that all texts in a collection will be loaded here? Those should be paginated, since they might be a lot.

jdamerow (changes requested):

annotations/templates/annotations/project_list.html
Girik1105: This will redirect the user to the text with `?project_id=` which is later used by other views to retrieve the project id

annotations/exceptions.py
```diff
@@ -9,16 +11,27 @@ def custom_403_handler(request, exception):
...
     ----------
     :class:`django.http.response.HttpResponse`
         Status 403.
     """
-    template = loader.get_template()
+    template = loader.get_template('annotations/forbidden_error_page.html')
+    
+    # Get custom error message from the PermissionDenied exception
+    # raise PermissionDenied("Custom message")
+    error_message = str(exception)
```
jdamerow: if `None` is passed as exception, this will make `error_message` `"None"`, won't it?

annotations/utils.py
```diff
@@ -64,4 +67,115 @@ def get_pagination_metadata(total_items, page, items_per_page):
...
+        order_direction = '-'
+    else:
+        order_field = order_by
+        order_direction = ''
+    
+    # Validate order field
+    if order_field not in allowed_fields:
+        order_field = default_field
+        order_direction = ''
+        order_by = default_field
```
jdamerow, line 107: there are two fields here that seem to hold the same value (`order_field`, `order_by`),  why are they both required?

annotations/utils.py
```diff
@@ -64,4 +67,115 @@ def get_pagination_metadata(total_items, page, items_per_page):
...
+    Returns
+    -------
+    dict
+        Dictionary containing order_by parameter and order_field to use in query
+    """
+    if allowed_fields is None:
+        allowed_fields = [default_field]
+        
+    # Get order_by from request, default to default_field
+    order_by = request.GET.get('order_by', default_field)
```
jdamerow: you probably want to make `'order_by'` also a parameter (with a default value), so it could be used for requests with a different parameter name.

annotations/views/project_views.py
```diff
@@ -8,60 +8,61 @@
...
-    request : `django.http.requests.HttpRequest`
-    project_id : int
-
-    Returns
-    ----------
-    :class:`django.http.response.HttpResponse`
     """
+    project = get_object_or_404(
+        _annotate_project_counts(TextCollection.objects),
+        pk=project_id
```
jdamerow, line 61: I'm so confused by this, doesn't `get_object_or_404` expect a class, not a queryset? A

annotations/views/project_views.py
```diff
@@ -8,60 +8,61 @@
...
     template = "annotations/project_details.html"
 
-    order_by = request.GET.get('order_by', 'title')
-    texts = project.texts.all().order_by(order_by)\
-                         .values('id', 'title', 'added', 'repository_source_id')
+    # Get ordering metadata
+    ordering = get_ordering_metadata(request, default_field='title', allowed_fields=['title', 'added'])
+
+    # Apply ordering
+    texts = project.texts.all().order_by(ordering['order_param'])\
```
jdamerow, line 74: same as in other place, we don't want to list all texts in a proejct on one page (what if it's multiple hundreds?).

annotations/views/repository_views.py
```diff
@@ -219,23 +221,46 @@ def repository_list(request):
 @citesphere_authenticated
 def repository_details(request, repository_id):
     template = "annotations/repository_details.html"
-    user = None if isinstance(request.user, AnonymousUser) else request.user
     repository = get_object_or_404(Repository, pk=repository_id)
+    user = None if isinstance(request.user, AnonymousUser) else request.user
+
+    # Fetch text collections (projects) owned by the current user
+    user_owned_collections = TextCollection.objects.filter(ownedBy=user)
```
jdamerow: is this used anywhere?

annotations/views/repository_views.py
```diff
@@ -272,7 +297,16 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
 
 
 @citesphere_authenticated
-def repository_text_import(request, repository_id, group_id, text_key):
+def repository_text_import(request, repository_id, group_id, text_key, project_id=None):
+
+    if not project_id:
+        return redirect(f"{reverse('list_projects')}?redirect_to_text_import=True&repository_id={repository_id}&group_id={group_id}&text_key={text_key}")
+    
+    print("project_id", project_id)
```
jdamerow: remove

annotations/views/rest_views.py
```diff
@@ -36,6 +36,55 @@
...
 
+# Custom permission class to check if user is collaborator or owner
+class IsCollaboratorOrOwner(IsAuthenticatedOrReadOnly):
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):
+            return False
+            
+        # Allow GET requests for authenticated users
+        if request.method in ['GET', 'HEAD', 'OPTIONS']:
+            return True
```
jdamerow, line 49: I'm confused, why would it matter if this is a GET or POST request?

jdamerow: Hm, ok, I think I slowly understand, I think it's a naming issue. The class name suggests this will just check if the user is collaborator and owner, not that it checks specific texts for edit permissions (which I think this is doing?). Either it should check in general if the user can see/edit the text or the class name needs to be way more specific. In either case, the class name needs to be changed though.

annotations/views/rest_views.py
```diff
@@ -36,6 +36,55 @@
...
+
+    def has_object_permission(self, request, view, obj):
+        # Allow GET requests for authenticated users
+        if request.method in ['GET', 'HEAD', 'OPTIONS']:
+            return True
+            
+        # Check if user is owner or collaborator of the text's collection
+        text = None
+        if hasattr(obj, 'occursIn'):
+            text = obj.occursIn
```
jdamerow: what is `obj`?

jdamerow (changes requested):

annotations/views/project_views.py
Girik1105: `get_object_or_404` can also accept a queryset. The utility function `_annotate_project_counts(TextCollection.objects)` returns an annotated queryset, where I have added fields like num_texts, num_relations, and num_collaborators. By passing this annotated queryset to get_object_or_404 ensures that these additional fields are included in the retrieved project object (mainly used fore rendering the project stats)

https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#get-object-or-404

annotations/models.py
```diff
@@ -411,7 +413,11 @@ class Text(models.Model):
     """
 
     def get_absolute_url(self):
-        return reverse('annotate', args=[self.id])
+        """Get the absolute URL for viewing/annotating this text."""
+        # Get the first project this text belongs to
+        project = self.partOf.first()
```
jdamerow: this doesn't seem right. Wouldn't that mean that if a text is in multiple projects of multiple users, a user might see a text in a project that is not their own?

annotations/views/project_views.py
jdamerow: hm, I think there needs to be a comment here, especially since it would be more obvious why you do that it here are many projects and you want to annotate them all, but this is only requesting one single one. It also doesn't help that it's the "TextCollection" class but the id is "project_id". So there should be an explanatory comment why this is used over just getting the counts directly from the requested project (the way it's written makes it a lot harder to understand what's going on).

jdamerow (changes requested):

PR #12, opened 2024-10-22
Story/VOGRE-8

rajvi-patel-22:

Use Conceptpower's json endpoint
https://diging.atlassian.net/browse/VOGRE-8

... describe here ...


jdamerow:
Please make sure to provide the ticket title if there is no description.

annotations/views/rest_views.py
```diff
@@ -596,66 +577,39 @@ def get_queryset(self, *args, **kwargs):
...
-                if authority_elem is not None:
-                    concept['authority'] = authority_elem.text.strip()
-
-                # Extract pos (if available)
-                pos_elem = concept_entry.find('skos:note', namespace)
-                if pos_elem is not None:
-                    concept['concept_type'] = pos_elem.text.strip()
-
-                return concept
+                    concept['authority'] = {'name': 'Unknown'}
```
jdamerow: the contents of this loop looks like it could be more DRY by extracting a method and reuse it here and in `search`?

jdamerow (changes requested):

annotations/views/rest_views.py
```diff
@@ -672,4 +593,28 @@ def _relabel(datum):
...
+    concept['pos'] = concept_entry.get('pos', '')
+    concept['concept_type'] = concept_entry.get('type','')
+
+    description = concept_entry.get('description', '')
+    try:
+        concept['description'] = json.loads(f'"{description}"')
+    except json.JSONDecodeError:
+        concept['description'] = description
+    
+    if concept['uri'].startswith(('http://www.digitalhps.org/', 'https://www.digitalhps.org/')):
```
jdamerow: those should be a setting in case this changes

jdamerow (changes requested):

annotations/views/rest_views.py
```diff
@@ -492,59 +493,23 @@ def search(self, request, **kwargs):
         if not q:
             return Response({'results': []})
         pos = request.GET.get('pos', None)
-        url = f"{settings.CONCEPTPOWER_ENDPOINT}ConceptLookup/{q}/{pos if pos else ''}"
-        response = requests.get(url, auth=(settings.CONCEPTPOWER_USERID, settings.CONCEPTPOWER_PASSWORD))
+        url = f"{settings.CONCEPTPOWER_ENDPOINT}ConceptSearch"
+        parameters = {
+            'word': q,
+            'pos': pos if pos else '',
```
jophals: So if `''` is passed for `pos` then this will result in a 400 error from ConceptPower. Passing `None` will actually return 'Any' pos

PR #13, opened 2024-10-29
Story/VOGRE-25

rajvi-patel-22:

 The prefix of the vogon app needs to be configurable. ([VOGRE-25](https://diging.atlassian.net/browse/VOGRE-25))
Right now, Vogon assumes it runs at the root of the server (e.g. localhost:8000/), but it needs to be configurable to run under e.g. localhost:8000/vogon (so the vogon part needs to be configurable).

Otherwise it can’t be deployed to the server. We have the same thing in DST.

... describe here ...

[VOGRE-25]: https://diging.atlassian.net/browse/VOGRE-25?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vogon/urls.py
```diff
@@ -79,71 +78,71 @@
...
 
-    re_path(r'^sandbox/(?P<text_id>[0-9]+)/$', conceptViews.sandbox, name='sandbox'),
+    re_path(rf'^{settings.APP_ROOT}sandbox/(?P<text_id>[0-9]+)/$', conceptViews.sandbox, name='sandbox'),
 
-    path('login/citesphere/', externalAccountViews.citesphere_login, name='citesphere_login'),
-    path('oauth/callback/citesphere/', externalAccountViews.citesphere_callback, name='citesphere_callback'),
-    path('citesphere/refresh/<int:repository_id>/', externalAccountViews.citesphere_refresh_token, name='citesphere_refresh_token'),
+    path(f'{settings.APP_ROOT}login/citesphere/', externalAccountViews.citesphere_login, name='citesphere_login'),
+    path(f'{settings.APP_ROOT}oauth/callback/citesphere/', externalAccountViews.citesphere_callback, name='citesphere_callback'),
+    path(f'{settings.APP_ROOT}citesphere/refresh/<int:repository_id>/', externalAccountViews.citesphere_refresh_token, name='citesphere_refresh_token'),
```
jdamerow: let's follow the approach from dst and prefix all urls in one go (not one in each url)
https://github.com/diging/dst/blob/develop/code/dst/urls.py

jdamerow (changes requested):

.env_app-example
```diff
@@ -10,4 +10,5 @@ CONCEPTPOWER_USERID=
 CONCEPTPOWER_PASSWORD=
 CONCEPTPOWER_ENDPOINT=
 CONCEPTPOWER_NAMESPACE=
-BASE_URL=http://localhost:8000/
\ No newline at end of file
+BASE_URL=http://localhost:8000/
+APP_ROOT=
```
jdamerow: there should be a comment that the `APP_ROOT` variable should end with `/`

jdamerow (changes requested):
- root needs to work with and with a slash at the end (`http://localhost:8000/vogon` does not work at the moment)

PR #15, opened 2024-11-05
story/VOGRE-7

Girik1105:

Get rid of celery and submit quaduples on request
Description

Currently, quadruples are submitted via Celery on a cron job. Celery should be completely removed. Instead a project should have a button that will submit quadrupes to Quadriga. This ticket includes multiple tasks:
Quadruples need to have a flag (if they don’t already) that indicates if they ahve been submitted or not
Quaduples need to be submitted to the new Quadriga 2.0
Only quadruples with complete relationships (in which concepts have been resolved and added to conceptpower) should be submitted.

[VOGRE-7](https://diging.atlassian.net/browse/VOGRE-7?atlOrigin=eyJpIjoiNjRlZWM4ZDUwM2Y1NGIzYmEwNWJiOGM3NDZlNDMyZjgiLCJwIjoiaiJ9)

[VOGRE-14](https://github.com/diging/vogon-resuscitate/pull/11)
  
Since VOGRE-14 has url changes which also affect `annotator `and `annotator factory` this PR will require changes once VOGRE-14 merged.

[VOGRE-7]: https://diging.atlassian.net/browse/VOGRE-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env_app-example
```diff
@@ -10,4 +10,6 @@ CONCEPTPOWER_USERID=
 CONCEPTPOWER_PASSWORD=
 CONCEPTPOWER_ENDPOINT=
 CONCEPTPOWER_NAMESPACE=
-BASE_URL=http://localhost:8000/
\ No newline at end of file
+BASE_URL=http://localhost:8000/
+QUADRIGA_ENDPOINT=
+QUADRIGA_COLLECTION_ID=
```
jdamerow: this is something that needs to be set for each project. Every project can have a different collection it submits quadruples to.

annotations/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+    - 'submitted': This is handled separately in the submission workflow and does not need to 
+                   be updated here.
+    
+    The purpose of this middleware is to automate the readiness check and ensure that 
+    RelationSets are always in the correct state before they can be submitted to Quadriga.
+    """
+
+    # Cache the last check time to reduce query frequency
+    last_check_time = None
+    check_interval = timedelta(minutes=5)  # Adjust the interval as needed
```
jdamerow: needs to be configurable through settings

annotations/models.py
```diff
@@ -787,12 +787,15 @@ class RelationSet(models.Model):
...
     """The text on which this RelationSet is based."""
 
-    pending = models.BooleanField(default=False)
-    """
-    A :class:`.RelationSet` is pending if it has been selected for submission,
-    but the submission process has not yet completed. The primary purpose of
-    this field is to prevent duplicate submissions.
-    """
+    STATUS_CHOICES = [
+        ('not_ready', 'Not Ready'),
```
jdamerow: the first value here should be defined in constants, so the constants can be used elsewhere (see below)

annotations/models.py
```diff
@@ -908,6 +911,17 @@ def ready(self):
...
 
+    def update_status(self):
+        """
+        Check if the RelationSet is ready and update the status accordingly.
+        """
+        if self.ready():  # Check readiness based on the concepts
+            if self.status != 'submitted':  # Avoid overriding submitted status
+                self.status = 'ready_to_submit'
+        else:
+            self.status = 'not_ready'
```
jdamerow: in line 920 and 922 the constants defined above should be sued

annotations/utils.py
```diff
@@ -37,6 +38,170 @@ def basepath(request):
...
+                    "expression": concept.label or "",
+                    "normalization": "",
+                    "formattedPointer": "",
+                    "format": ""
+                }
+            ]
+        },
+        "context": {
+            "creator": user.username,
+            "creationTime": timezone.now().strftime('%Y-%m-%d'),
```
jdamerow: this should be the time an appellation was created

annotations/utils.py
```diff
@@ -37,6 +38,170 @@ def basepath(request):
...
+            "creationPlace": "phoenix",
```
jdamerow: let's leave that empty for now

annotations/utils.py
```diff
@@ -37,6 +38,170 @@ def basepath(request):
...
+            "sourceUri": concept.authority if hasattr(concept, 'authority') else ""
```
jdamerow: this needs to be the the uri of the text that has been annotated, so for now let's use the the url of the annotated text

jdamerow: the above comments apply to all contexts

annotations/utils.py
```diff
@@ -37,6 +38,170 @@ def basepath(request):
...
+                    "creationPlace": "phoenix",
+                    "sourceUri": ""
+                }
+            },
+            "nodes": nodes,
+            "edges": edges
+        }
+    }
+
+def generate_graph_data(relationset, user):
```
jdamerow: is this the same function as above?

annotations/views/rest_views.py
```diff
@@ -328,6 +331,63 @@ def get_queryset(self, *args, **kwargs):
...
+            relationset = RelationSet.objects.get(pk=pk)
+
+        except RelationSet.DoesNotExist:
+            return Response({'error': 'RelationSet not found.'}, status=status.HTTP_404_NOT_FOUND)
+
+        if relationset.createdBy != user:
+            return Response({'error': 'You are not authorized to submit this RelationSet.'},
+                            status=status.HTTP_403_FORBIDDEN)
+
+        if relationset.status != 'ready_to_submit':
```
jdamerow: use constant here

jdamerow (changes requested):

jdamerow:
Also, there are conflicts

annotations/middleware.py
jdamerow, line 1: I don't think this is the right approach. It is likely to cause memory issues once there are sufficiently many relationsets. It also seems unnecessary. The only time we care about the readiness of a relation to be submitted is when we want to submit them or maybe when we look at a text. So it really only needs to be checked when the page shows all relations, and then only the relations that are relevant (not all of them).

annotations/utils.py
```diff
@@ -37,6 +38,114 @@ def basepath(request):
...
+                    "creationPlace": "",
+                    "sourceUri": source_uri
+                }
+            },
+            "nodes": nodes,
+            "edges": edges
+        }
+    }
+
+
```
jdamerow: these two functions are specific to the quadriga api, aren't they? They should not be in the annotations util file.

annotations/views/rest_views.py
jdamerow: this still

jdamerow: and everywhere else where these strings are used

concepts/tasks.py
```diff
@@ -14,23 +14,24 @@
...
+    # except Concept.DoesNotExist:
+    #     return
+
+    # try:
+    #     manager.resolve()
+    # except ConceptLifecycleException as E:
+    #     logger.debug("Resolve concept failed: %s" % str(E))
+    #     return 
+    # logger.debug("Resolved concept %s" % manager.instance.uri)
+    pass
```
jdamerow: remove commented out code

vogon/__init__.py
```diff
@@ -1,5 +1,4 @@
 from __future__ import absolute_import
 
 # This will make sure the app is always imported when
-# Django starts so that shared_task will use this app.
-from .celery import app as celery_app
+# Django starts so that shared_task will use this app.
```
jdamerow: are these comments still valid?

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

annotations/middleware.py
Girik1105: > I don't think this is the right approach. It is likely to cause memory issues once there are sufficiently many relationsets. It also seems unnecessary. The only time we care about the readiness of a relation to be submitted is when we want to submit them or maybe when we look at a text. So it really only needs to be checked when the page shows all relations, and then only the relations that are relevant (not all of them).

If we only show the relations that can be submitted, then the user will not be able to see all relations that they have created on the text. Can we still show the relations that cannot be submitted with the disabled checkbox? I'll change the way we check for ready relations.

jdamerow:
Resolve conflicts please

.env_app-example
```diff
@@ -15,3 +15,5 @@ export BASE_URL="http://localhost:8000/"
 # Make sure APP_ROOT variable ends with a /
 # this is the prefix that vogon will be deployed under (e.g. "vogon-2" for deployment at my-server.org/vogon-2)
 export APP_ROOT=
+
+export QUADRIGA_ENDPOINT=
```
jdamerow, line 21: this needs to have an example (as a comment), e.g http://quadriga.instance/quadriga/...whatever else needs to be here/

annotations/annotators.py
```diff
@@ -50,15 +50,9 @@ def my_view(request, text_id):
...
-
-import requests
 from django.shortcuts import get_object_or_404, render
 from django.http import Http404
-from annotations.tasks import tokenize
-from annotations.utils import basepath
-from annotations.models import TextCollection, VogonUserDefaultProject, Text
-from urllib.parse import urlparse
-import chardet
+from annotations.models import TextCollection, Text
```
jdamerow: is this needed here?

annotations/quadriga.py
```diff
@@ -275,3 +253,112 @@ def parse_response(raw_response):
...
+def build_concept_node(concept, user, creation_time, source_uri):
+    """
+    Helper function to build a concept node dictionary.
+    """
+    return {
+        "label": concept.label or "",
+        "metadata": {
+            "type": "appellation_event",
+            "interpretation": concept.uri,
+            "termParts": [
```
jdamerow: there should be one term part for every piece of annotated text in an appellation (I think currently there can only be one term part as vogon doesn't allow highlighting multiple parts that are not together for one appellation). Position needs to be the position of the first character of the term part.

annotations/quadriga.py
```diff
@@ -275,3 +253,112 @@ def parse_response(raw_response):
...
+                {
+                    "position": 1,
+                    "expression": concept.label or "",
```
jdamerow: expresssion is the annotated text

annotations/quadriga.py
```diff
@@ -275,3 +253,112 @@ def parse_response(raw_response):
...
+        # Extract the creation time from the appellation
+        creation_time = relationset.occursIn.created
+
+        if subject and subject.uri not in nodes:
+            subject_id = get_node_id()
+            nodes[subject_id] = build_concept_node(subject, user, creation_time, source_uri)
+
+        if obj and obj.uri not in nodes:
+            obj_id = get_node_id()
+            nodes[obj_id] = build_concept_node(obj, user, creation_time, source_uri)
```
jdamerow: what about nested relations, in which subject or object are a relation?

annotations/views/rest_views.py
```diff
@@ -374,6 +381,70 @@ def get_queryset(self, *args, **kwargs):
...
+                'Content-Type': 'application/json',
+            }
+
+            collection_id = project.quadriga_id
+            endpoint = f"{settings.QUADRIGA_ENDPOINT}/api/v1/collection/{collection_id}/network/add/"
+
+            graph_data = generate_graph_data(relationset, user)
+
+            response = requests.post(endpoint, json=graph_data, headers=headers)
+            response.raise_for_status()
```
jdamerow: all the quadriga related things should happen in a quadriga file or class

vogon/__init__.py
jdamerow: ?

jdamerow (changes requested):

annotations/models.py
```diff
@@ -785,6 +785,9 @@ class RelationSet(models.Model):
     A :class:`.RelationSet` organizes :class:`.Relation`\s into complete
     statements.
     """
+    STATUS_NOT_READY = 'not_ready'
+    STATUS_READY_TO_SUBMIT = 'ready_to_submit'
+    STATUS_SUBMITTED = 'submitted'
```
jdamerow: let's keep those with the chioce field they are used in

annotations/quadriga.py
```diff
@@ -275,3 +257,150 @@ def parse_response(raw_response):
...
+    
+    term_parts = []
+    for appellation in appellations:
+        term_parts.append({
+            "position": appellation.startPos if appellation.startPos else 0,
+            "expression": appellation.stringRep or "",
+            "normalization": "",
+            "formattedPointer": "",
+            "format": ""
+        })
```
jdamerow: I'm confused, I don't think this is correct. per appellation there might be more than one term part, but those term parts are not coming from more than one appellation. 
appellation -has -> concept
appellation -has -> term parts

but not all term parts for all appellations should be collapsed into one node. Does that make sense?

annotations/quadriga.py
```diff
@@ -275,3 +257,150 @@ def parse_response(raw_response):
...
+    
+    return {
+        "label": concept.label or "",
+        "metadata": {
+            "type": "appellation_event", 
+            "interpretation": concept.uri,
+            "termParts": term_parts
```
jdamerow, line 292: per appellation event, there should only be the term parts belonging to that appellation (not all appellations with the same concept).

annotations/quadriga.py
jdamerow: this still I think

annotations/views/rest_views.py
```diff
@@ -374,6 +381,49 @@ def get_queryset(self, *args, **kwargs):
...
+
+        try:
+            submit_to_quadriga(relationset, user, project)
+            return Response({'success': 'Quadruples submitted successfully.'}, status=status.HTTP_200_OK)
+
+        except CitesphereAccount.DoesNotExist:
+            return Response({'error': 'No Citesphere account found.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        except requests.RequestException as e:
+            print("ERROR", e)
```
jdamerow: needs to use logger

jdamerow (changes requested):

annotations/models.py
```diff
@@ -929,6 +938,20 @@ def ready(self):
...
 
+    def update_status(self):
+        """
+        Check if the RelationSet is ready and update the status accordingly.
+        """
+        if self.ready():  # Check readiness based on the concepts
+            if self.status != self.STATUS_SUBMITTED:  # Avoid overriding submitted status
+                self.status = self.STATUS_READY_TO_SUBMIT
+                self.submitted = False
+        else:
```
jdamerow, line 952: doesn't this set `status` and `submitted` even if those have already the correct values? That seems unnecessary.

annotations/quadriga.py
```diff
@@ -275,3 +257,218 @@ def parse_response(raw_response):
...
+    """
+    Build a node dictionary for an appellation event.
+    Each appellation gets its own node even if it points to the same concept.
+    The node’s termParts come only from the given appellation.
+
+      - "metadata.interpretation" is the concept’s label.
+      - "context.sourceUri" is the URL of the concept source (from appellation.interpretation.master.uri)
+    """
+    term_parts = []
+    pos = appellation.startPos if appellation.startPos is not None else 0
```
jdamerow: hm, no, 0 and None is different. if there is no start position then there is probably no word actually being annotated. in that case we probably want to keep pos None (or null or whatever the json value is), as 0 would indicate the annotated word starts at position 0.

annotations/quadriga.py
```diff
@@ -275,3 +257,218 @@ def parse_response(raw_response):
...
+        "expression": exp,
+        "normalization": "",
+        "formattedPointer": "",
+        "format": ""
+    })
+    
+    # concept's label for the interpretation.
+    interpretation_label = appellation.interpretation.label if appellation.interpretation.label else ""
+    # concept source URL from the master attribute if available.
+    if hasattr(appellation.interpretation, 'master') and hasattr(appellation.interpretation.master, 'uri'):
```
jdamerow, line 282: what is "master"?

annotations/quadriga.py
```diff
@@ -275,3 +257,218 @@ def parse_response(raw_response):
...
+        "label": interpretation_label,
+        "metadata": {
+            "type": "appellation_event",
+            "interpretation": interpretation_label,
+            "termParts": term_parts
+        },
+        "context": {
+            "creator": user.username,
+            "creationTime": creation_time.strftime('%Y-%m-%d'),
+            "creationPlace": "phoenix",
```
jdamerow: The value of this one should be configurable.

annotations/quadriga.py
```diff
@@ -275,3 +257,218 @@ def parse_response(raw_response):
...
+            "sourceUri": concept_source_url
```
jdamerow: this should be the uri of the text

annotations/quadriga.py
```diff
@@ -275,3 +257,218 @@ def parse_response(raw_response):
...
+    
+    # default mapping from the top-level relation's roles.
+    top_subj = top_relation.source_content_object
+    top_pred = top_relation.predicate
+    top_obj = top_relation.object_content_object
+    default_mapping = {
+        "subject": {"type": "REF", "reference": node_mapping.get(f"app-{top_subj.id}-{top_subj.created.isoformat()}", "")},
+        "predicate": {"type": "URI", "uri": top_pred.interpretation.label, "label": top_pred.interpretation.label},
+        "object": {"type": "REF", "reference": node_mapping.get(f"app-{top_obj.id}-{top_obj.created.isoformat()}", "")}
+    }
```
jdamerow, line 492: not sure I understand what this is doing, but the default relation would need to be configured in the template that what used to create a relation.

jdamerow (changes requested):

annotations/quadriga.py
Girik1105: This is used to get the highest-level merge target, its a concept [Model function](https://github.com/diging/vogon-resuscitate/blob/story/VOGRE-7/concepts/models.py#L94). This would help us get the concept's uri if available or else it will fallback to the text's uri which is the vogon uri (Example: urn:repository:1:item:ZVIGSRR8:file:FILEs8vb4PIqOQB7)

annotations/quadriga.py
Girik1105: This is the concept node that is why I added the concept's uri instead of the text uri, I add the text uri in the graph data ([implementation](https://github.com/diging/vogon-resuscitate/blob/story/VOGRE-7/annotations/quadriga.py#L418)). Should I replace this also as the text uri instead?

annotations/models.py
```diff
@@ -929,6 +938,31 @@ def ready(self):
...
+        Check if the RelationSet is ready and update the status accordingly.
+        """
+        changed = False
+        
+        if self.ready():  # Check readiness based on the concepts
+            if self.status != self.STATUS_SUBMITTED:  # Avoid overriding submitted status
+                if self.status != self.STATUS_READY_TO_SUBMIT:
+                    self.status = self.STATUS_READY_TO_SUBMIT
+                    changed = True
+                if self.submitted:
```
jdamerow: how could that be set to true? if the status is not submitted, then this should be false, right?

annotations/models.py
```diff
@@ -929,6 +938,31 @@ def ready(self):
...
+                    self.submitted = False
+                    changed = True
```
jdamerow:
```suggestion
               self.status = self.STATUS_READY_TO_SUBMIT if self.status != self.STATUS_READY_TO_SUBMIT
               self.submitted = False
               self.save()
```

annotations/models.py
```diff
@@ -929,6 +938,31 @@ def ready(self):
...
+        else:
+            if self.status != self.STATUS_NOT_READY:
```
jdamerow: is this check necessary? if the object is not ready, then it should be ignored, right?

annotations/quadriga.py
jdamerow: But that's not a text uri, it's a concept?

annotations/quadriga.py
jdamerow: yes, source uri should always be the uri of the annotated text

annotations/quadriga.py
```diff
@@ -275,3 +257,280 @@ def parse_response(raw_response):
...
+    # 3. object - The entity that the subject relates to
+    #
+    # For Quadriga, this default mapping:
+    # - Provides an entry point to understand the graph
+    # - Identifies which relationship is most important
+    # - Allows immediate access to key parts without traversing the whole graph
+    #############################################################################
+    
+    # Initialize an empty default mapping dictionary
+    default_mapping = {}
```
jdamerow, line 422: the default mapping should be coming from the template used to create a relation

annotations/views/rest_views.py
```diff
@@ -374,6 +381,49 @@ def get_queryset(self, *args, **kwargs):
             queryset = queryset.filter(project_id=project_id)
 
         return queryset.order_by('-created')
+    
+    @action(detail=False, methods=['post'], permission_classes=[IsAuthenticated], url_name='submit')
+    def submit(self, request):
+
+        user = request.user
+        pk = request.data.get('pk')
```
jdamerow: bad variable name. pk of what?

jdamerow (changes requested):

annotations/forms.py
```diff
@@ -246,16 +247,70 @@ class RelationTemplateForm(forms.ModelForm):
...
+            'class': 'form-control input-sm relation-node-input',
+            'placeholder': 'First relation node ID'
+        }))
+    relation_node_2 = forms.CharField(required=False, widget=forms.TextInput(attrs={
+            'class': 'form-control input-sm relation-node-input',
+            'placeholder': 'Second relation node ID'
+        }))
+    concept_behavior = forms.CharField(required=False, widget=forms.TextInput(attrs={
+            'class': 'form-control input-sm relation-node-input',
+            'placeholder': 'How does the concept behave? (e.g. teaches, influences)'
```
jdamerow, line 268: what is this for? what's the concept behavior?

annotations/forms.py
```diff
@@ -246,16 +247,70 @@ class RelationTemplateForm(forms.ModelForm):
...
+        node1 = cleaned_data.pop('relation_node_1', None)
+        node2 = cleaned_data.pop('relation_node_2', None)
+        behavior = cleaned_data.pop('concept_behavior', None)
+        predicate_uri = cleaned_data.pop('predicate_concept', None)
+        
+        # Check if using alternative input method
+        if any([node1, node2, behavior, predicate_uri]):
+            # Validate all fields are present for alternative method
+            if not all([node1, node2, behavior, predicate_uri]):
+                raise forms.ValidationError("All fields are required when using relation nodes input")
```
jdamerow, line 290: I don't understand what this message means

annotations/models.py
jdamerow: this still

annotations/quadriga.py
jdamerow: ?

annotations/quadriga.py
```diff
@@ -275,3 +257,303 @@ def parse_response(raw_response):
...
+    
+    # Get the template part from the relationset's template
+    template_part = relationset.template.template_parts.first()
+    
+    # Get the subject node from the relation
+    subject_node = None
+    if template_part.source_node_type == 'CO':  # Specific concept
```
jdamerow, line 429: For all these strings, the appropriate constant from models should be used.

annotations/quadriga.py
```diff
@@ -275,3 +257,303 @@ def parse_response(raw_response):
...
+            "reference": node_mapping.get(subj_key, "0"),
+        },
+        "predicate": {
+            "type": "REF" if is_alternative_format else "URI",
+            "uri": alt_format_uri if is_alternative_format else predicate_node.interpretation.master.uri,
+        },
+        "object": {
+            "type": "REF",
+            "reference": node_mapping.get(obj_key, "0"),
+        }
```
jdamerow, line 491: all three parts of the default mapping could be either a specific concept or a reference to a node in the graph.

jdamerow (changes requested):

annotations/forms.py
Girik1105: This is the predicate, I'll change the placeholder

Girik1105:
https://github.com/diging/vogon-resuscitate/pull/15#discussion_r2028907535

Ive set the function so works by first checking if the RelationSet is ready for submission using the ready() method, which checks is the concepts are resolved and ready to be submitted. If the RelationSet is ready and hasn't been submitted yet, the method updates its status to STATUS_READY_TO_SUBMIT and ensures the submitted flag is set to False, indicating it's prepared but not yet submitted. The else part verifies if the RelationSet isn't ready, it sets the status to STATUS_NOT_READY and also marks it as not submitted.

PR #14, opened 2024-10-31
story/VOGRE-12

Girik1105:
# Guidelines for Pull Requests

If you haven't yet read our code review guidelines, please do so, You can find them [here](https://diging.atlassian.net/wiki/spaces/DIGING/pages/2256076801/Code+Review+Guidelines).

Please confirm the following by adding an x for each item (turn `[ ]` into `[x]`).

- [x] I have removed all code style changes that are not necessary (e.g. changing blanks across the whole file that don’t need to be changed, adding empty lines in parts other than your own code)
- [x] I am not making any changes to files that don’t have any effect (e.g. imports added that don’t need to be added)
- [x] I do not have any sysout statements in my code or commented out code that isn’t needed anymore
- [x] I am not reformatting any files in the wrong format or without cause. 
- [x] I am not changing file encoding or line endings to something else than UTF-8, LF
- [x] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed
- [x] I have added Javadoc/documentation where appropriate
- [x] I have added test cases where appropriate
- [x] I have explained any part of my code/implementation decisions that is not be self-explanatory

## Please provide a brief description of your ticket 
This ticket requires to show the user all files associated with a specific text in citesphere and allow the user to import any inidivudal file the user would like.

# Multiple texts for a Citesphere entry

Description:
Citesphere entries can have more than one text attached to them (right now I think the first one will be used?).
The text table should list all the file names that are attached to a record, and then instead of an “Import”  button, there should be an import icon next to each file.

https://diging.atlassian.net/browse/VOGRE-12?atlOrigin=eyJpIjoiNWJjMWNmMGJiYWYyNDdmMzkyYmIyZWMyMGZiMTA0MTEiLCJwIjoiaiJ9

## Are there any other pull requests that this one depends on?

[VOGRE-14](https://github.com/diging/vogon-resuscitate/pull/11)

## Anything else the reviewer needs to know?

I Updated the `repository_text_import` view to support file-specific imports by using `file_id` as part of the unique identifier (`uri`). This change ensures that each file import, even for the same `text_key`, creates or updates a distinct `Text` object. The view now returns `JsonResponse`, providing error messages and a `redirect_url` for successful imports, since now all files are being rendered via AJAX. 

I might have to make some changes to this PR when VOGRE-14 gets merged into develop since both of these tickets have changes on the `repository_text_import` view

[...]

PR #18, opened 2024-11-07
bug/VOGRE-27

Girik1105:

Fix issues with multiple citesphere intances

Description
find by user and repository
giles needs to be specific for citespher account

[VOGRE-27](https://diging.atlassian.net/browse/VOGRE-27?atlOrigin=eyJpIjoiNzFhOTY2NTY0MDFiNDc1MmJjZWY2ZDkzOGU1MmQ3MWMiLCJwIjoiaiJ9)

[VOGRE-12](https://github.com/diging/vogon-resuscitate/pull/14)
  
I added better error handling for Giles Uploads since even when there isn't valid text it would still create text instances for it.

[VOGRE-27]: https://diging.atlassian.net/browse/VOGRE-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


external_accounts/utils.py
```diff
@@ -45,14 +45,15 @@ def get_giles_document_details(user, file_id):
...
+        citesphere_account = CitesphereAccount.objects.get(user=user, repository=repository)
         if not citesphere_account:
             return None
         token = citesphere_account.access_token
         headers = {'Authorization': f'Bearer {token}'}
-        url = f"{settings.GILES_ENDPOINT}api/v2/resources/files/{file_id}/content/"
+        url = f"{repository.giles_endpoint}/api/v2/resources/files/{file_id}/content/"
 
         response = requests.get(url, headers=headers)
+        print(response)
```
jdamerow: remove

repository/managers.py
```diff
@@ -137,33 +138,32 @@ def item(self, groupId, itemId):
...
+            if not giles_uploads:
+                raise GilesUploadError("No Giles uploads available for this item.")
+            
+            giles_details = []
+            extracted_text = giles_uploads[0].get('extractedText', {})
+
+            if extracted_text and extracted_text.get('content-type') == 'text/plain':
+                extracted_text_data = get_giles_document_details(self.user, extracted_text.get('id'), repository)
+                item_data['item']['text'] = extracted_text_data
+            elif giles_uploads[0].get('pages'):
```
jdamerow: hm, I don't think we want this. if there are individual pages, then the count would be off when they get appended together. I think for now, we ignore this. If there is no extracted text, then we just can't annotate the text. However, if a text file has been submitted to Giles, I don't think the text comes back as extracted text but as uploaded file. Have you check that?

jdamerow: in that case, we want to annotate the uploaded text.

jdamerow (changes requested):

repository/managers.py
Girik1105: When I uploaded a file, I did get `extractedText` back along with `uploadedFile`, so can I keep using extractedText to get the text?

```
  "uploadedFile": {
      "@type": "GilesFile",
      "id": "",
      "filename": "abcd.pdf",
      "url": "",
      "size": ,
      "processor": ,
      "content-type": "application/pdf"
  },
  "extractedText": {
      "@type": "GilesFile",
      "id": "",
      "filename": "abcd.pdf.txt",
      "url": "",
      "size": 6681,
      "processor": null,
      "content-type": "text/plain"
}
```

annotations/views/repository_views.py
```diff
@@ -307,9 +307,16 @@ def repository_text_import(request, repository_id, group_id, text_key, project_i
...
     try:
-        result = manager.item(group_id, text_key)
+        result = manager.item(group_id, text_key, repository)
     except IOError:
-        return render(request, 'annotations/repository_ioerror.html', {}, status=500)
+        return render(request, 'annotations/repository_ioerror.html', {'error': "IOError occurred while accessing the repository."}, status=500)
+    except GilesUploadError as e:
+        return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
+    except GilesTextExtractionError as e:
+        return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
```
jdamerow: all these errors need to be logged

annotations/templates/annotations/repository_ioerror.html
```diff
@@ -5,10 +5,10 @@
 {% block main %}
 <div class="container">
     <div class="h1 text-danger">Whoops! <i class="fa fa-frown-o" aria-hidden="true"></i></div>
+    <p class="text-warning">{{ error }}</p>
     <p class="text-warning">
-        There was a problem communicating with the remote repository that contains
-        this content. Please go back, and try again. If this problem persists,
-        please contact an <a href="mailto:erick.peirson@asu.edu">administrator</a>.
+        Please go back, and try again. If this problem persists,
+        please contact an <a href="mailto:jdamerow@asu.edu">administrator</a>.
```
jdamerow: let's remove the email address

external_accounts/giles.py
```diff
@@ -0,0 +1,163 @@
...
+        headers = {'Authorization': f'Bearer {self.access_token}'}
+        url = f"{self.base_url}/api/v2/resources/files/upload/{upload_id}"
+        
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+        
+        return response.json()[0] # API returns list with single item
+    
+
+def get_giles_document_details(user, file_id, repository):
```
jdamerow: let's put those functions into a class that is initialized with user and repository. we'll probably need it more eventually and this way we don't have to keep passing user and repository.

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

annotations/views/repository_views.py
jdamerow: this still

repository/managers.py
jdamerow: if you upload a txt file, you should not get an extracted text as there would be no extracted text

external_accounts/giles.py
```diff
@@ -0,0 +1,163 @@
...
+        extracted_text = upload_info.get("extractedText")
+        if extracted_text and extracted_text.get("content-type") == "text/plain":
+            text_content = giles.get_file_content(extracted_text.get("id"))
+            if not text_content or '\x00' in text_content:
+                return {"status": "error", "message": "The text you are trying to import contains invalid text content containing null characters This may be due to Giles processing."}
+            return {
+                "status": "complete",
+                "extracted_text" : text_content
+            }
+        else:
```
jdamerow: see my comment; for txt files this will fail I believe

external_accounts/giles.py
```diff
@@ -0,0 +1,163 @@
...
+
+        upload_info = giles.get_upload_details(upload_id)  # API returns list with single item
+        
+        if upload_info.get("documentStatus") != "COMPLETE":
+            return {"status": "processing", "message": "Files are still being processed in Giles, Please check back later!"}
+
+        extracted_text = upload_info.get("extractedText")
+        if extracted_text and extracted_text.get("content-type") == "text/plain":
+            text_content = giles.get_file_content(extracted_text.get("id"))
+            if not text_content or '\x00' in text_content:
```
jdamerow: when does this happen?

repository/managers.py
```diff
@@ -231,7 +232,23 @@ def item(self, groupId, itemId, fileId):
         }
 
         # Extract Giles uploads and their text if available
-        text = get_giles_document_details(self.user, fileId)
+        try:
+            text = get_giles_document_details(self.user, fileId, repository)
+            if text is None:
+                raise GilesTextExtractionError("Failed to retrieve text content from Giles, the file does not exist.")
+        except requests.RequestException as e:
+            raise CitesphereAPIError(message=f"Error accessing Giles API: {str(e)}",
```
jdamerow: why does this raise a CitesphereAPIError if the error is with Giles?

repository/managers.py
```diff
@@ -231,7 +232,23 @@ def item(self, groupId, itemId, fileId):
...
+        except requests.RequestException as e:
+            raise CitesphereAPIError(message=f"Error accessing Giles API: {str(e)}", 
+                                   error_code="GILES_API_ERROR",
+                                   details=str(e))
+        except ValueError as e:
+            raise CitesphereAPIError(message=f"Authentication error: {str(e)}",
+                                   error_code="AUTH_ERROR", 
+                                   details=str(e))
+        except Exception as e:
+            raise CitesphereAPIError(message=f"Unexpected error retrieving Giles document: {str(e)}",
```
jdamerow: same for all of these really

jdamerow (changes requested):

external_accounts/giles.py
Girik1105: image

Error caused on link: http://localhost:8000/vogon/repository/1/group/2256709/text/9TSD5RVD/file/FILENVkfPsB980sn/project/1/

Some texts give me this error so I added a check for NUL characters

annotations/views/repository_views.py
```diff
@@ -345,10 +346,18 @@ def repository_text_import(request, repository_id, group_id, text_key, file_id,
...
     try:
-        result = manager.item(group_id, text_key, file_id)
-    except IOError as e:
-        logger.error(f"Error accessing repository: {str(e)}")
-        return render(request, 'annotations/repository_ioerror.html', {'error': 'There was an error accessing the repository.'}, status=500)
+        result = manager.item(group_id, text_key, file_id, repository)
+    except IOError:
+        return render(request, 'annotations/repository_ioerror.html', {'error': "IOError occurred while accessing the repository."}, status=500)
+    except GilesUploadError as e:
+        logger.error(f"GilesUploadError: {str(e)}")
```
jdamerow: I don't think this prints the whole stack trace, but we want it. Or at least more info than the error message.

annotations/views/repository_views.py
```diff
@@ -345,10 +346,18 @@ def repository_text_import(request, repository_id, group_id, text_key, file_id,
...
+        return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
+    except GilesTextExtractionError as e:
+        logger.error(f"GilesTextExtractionError: {str(e)}")
```
jdamerow: same here

annotations/views/repository_views.py
```diff
@@ -345,10 +346,18 @@ def repository_text_import(request, repository_id, group_id, text_key, file_id,
...
+        return render(request, 'annotations/repository_ioerror.html', {'error': str(e)}, status=500)
+    except Exception as e:
+        # Catch any other exceptions
+        return render(request, 'annotations/repository_ioerror.html', {'error': "An unexpected error occurred: " + str(e)}, status=500)
```
jdamerow: same here, log error message

external_accounts/giles.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        if not self.access_token:
+            raise ValueError("User must authenticate with Citesphere before making API calls")
+            
+        headers = {'Authorization': f'Bearer {self.access_token}'}
+        url = f"{self.base_url}/api/v2/resources/files/{file_id}/content/"
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+        content = response.text
+        if '\x00' in content:
```
jdamerow, line 52: this needs a comment why this is here

external_accounts/giles.py
jdamerow: this should probably also go in the giles class

repository/managers.py
```diff
@@ -25,6 +28,7 @@ def _get_headers(self):
         try:
             return auth.citesphere_auth(self.user, self.repository)
         except Exception as e:
+            logger.error(f"Authentication failed: {str(e)}")
```
jdamerow: like everywhere, we want to log the stack trace

repository/managers.py
```diff
@@ -211,12 +222,18 @@ def item(self, groupId, itemId, fileId):
...
+            GilesUploadError: When there's an issue with Giles uploads
+            GilesTextExtractionError: When text content cannot be extracted
         """
         # Fetch item details using CitesphereAPIv1
-        item_data = self.api.get_item_details(groupId, itemId)
+        try:
+            item_data = self.api.get_item_details(groupId, itemId)
+        except CitesphereAPIError as e:
+            logger.error(f"Failed to fetch item details: {str(e)}")
+            raise
```
jdamerow: exceptions should be handled where they are caught. So no logging here if you just pass on the exception (in fac you can remove the whole try except block)

jdamerow (changes requested):
across the board, we want to either handle or raise an exception not both. logging should happen where the exception is caught, not where it is raised.

repository/managers.py
```diff
@@ -25,6 +28,8 @@ def _get_headers(self):
         try:
             return auth.citesphere_auth(self.user, self.repository)
         except Exception as e:
+            error_trace = traceback.format_exc()
+            logger.error(f"Error:\n{error_trace}")
```
jdamerow: I think I've made this comment before somewhere, either handle or throw an exception. Since it is thrown here, nothing else needs to be done. It should be logged where it is being handled. (same for all the other cases).

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

PR #23, opened 2024-11-14
story/VOGRE-19

rajvi-patel-22:

https://diging.atlassian.net/browse/VOGRE-19

- Fixed date appellation sidebar
    - Issue: Date Appellations were showing both date and text appellations
- Fixed select button for relation
    - Issue: Relation list item was not getting highlighted when we select it

### DIFF
--- a/annotations/static/annotations/js/annotators/appellationlist.js
+++ b/annotations/static/annotations/js/annotators/appellationlist.js
@@ -217,6 +217,7 @@ AppellationList = {
             // Replace an array prop wholesale doesn't seem to trigger a
             //  DOM update in the v-for binding, but a push() does; so we'll
             //  just push the appellations that aren't already in the array.
+            this.current_appellations = [];
             var current_ids = this.current_appellations.map(function (elem) {
                 return elem.id;
             });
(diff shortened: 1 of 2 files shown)


jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -217,6 +217,7 @@ AppellationList = {
             // Replace an array prop wholesale doesn't seem to trigger a
             //  DOM update in the v-for binding, but a push() does; so we'll
             //  just push the appellations that aren't already in the array.
+            this.current_appellations = [];
```
jdamerow: I'm confused, wouldn't this make the next line useless? Doesn't the next line create a list of ids of current appellations, which now would always be an empty list?

rajvi-patel-22: That's right. we don't need those lines.  Just `this.current_appellations = this.appellations;` will work. 
Can you confirm if the functionality of the buttons mentioned below is correct?
Screenshot 2024-12-06 at 10 06 22 AM

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

PR #25, opened 2024-11-15
story/VOGRE-16

rajvi-patel-22:

Fix Docker setup: [VOGRE-16](https://diging.atlassian.net/browse/VOGRE-16)
Vogon needs to be run through supervisor

[VOGRE-16]: https://diging.atlassian.net/browse/VOGRE-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


startup-prod.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+BASH_ENV=.env_app
```
jdamerow: Is this actually working? It stopped working for me for some reason.

supervisord.conf
```diff
@@ -0,0 +1,11 @@
+[supervisord]
+nodaemon=true
+
+[program:vogon]
+command=/vogon/startup.sh
```
jdamerow: this does not seem right. That's the dev startup script, right? So it will still run the dev server. Instead Django should be run with Guniconr and supervisor would start gunicorn. It should look somewhat like this probably:
https://github.com/diging/amphora-docker/tree/main/scripts
(without the celery part and there might be updates to it). Maybe google gunicorn, django, and supervisor and see if you can find an update ot date example.

jdamerow (changes requested):

startup-prod.sh
rajvi-patel-22: I guess there was an extra space after the `APP_ROOT` variable, which I have fixed:
`APP_ROOT= `-> `APP_ROOT=`

jdamerow: Let's use the line from the updated development version:
`source .env_app`

vogon-gunicorn.sh
```diff
@@ -0,0 +1,24 @@
...
+export PYTHONPATH=$DJANGODIR:$PYTHONPATH
+
+# Start your Django Unicorn
+# Programs meant to be run under supervisor should not daemonize themselves (do not use --daemon)
+exec gunicorn ${DJANGO_WSGI_MODULE}:application \
+  --name $NAME \
+  --workers $NUM_WORKERS \
+  --bind=0.0.0.0:8000 \
+  --log-level=info \
+  --log-file=-
```
jdamerow: log-file should log to a file.

startup-prod.sh
```diff
@@ -0,0 +1,8 @@
...
+mkdir -p /vogon/logs
+python -m pip install -r requirements.txt
+python manage.py migrate
+python manage.py collectstatic --noinput
+exec supervisord -c /etc/supervisor/conf.d/supervisord.conf
```
jdamerow: supervisor should be installed as a service so you can start it as such (`systemctl start supervisor` I believe)

jdamerow (changes requested):

jdamerow:
also there are conflicts

jdamerow (changes requested):
Please compare with https://github.com/diging/edrop-connector and set up similarly.

PR #28, opened 2024-12-03
story/VOGRE-23

Girik1105:

There should be a user type: Vogon Admin

Description
Vogon admins should have access to the templates (and can edit/add/delete them), but not the Django admin site.

[VOGRE-23](https://diging.atlassian.net/browse/VOGRE-23?atlOrigin=eyJpIjoiNmM2M2ZhMTc3MmZiNDc1MGFjOTQ0ZjdmZTg0ZGU3NTEiLCJwIjoiaiJ9)

I have added a panel called Vogon Admin Panel (can be access through the dropdown) which is only available to super users where they can browse through users in the application and toggle Vogon Admin status.

[VOGRE-23]: https://diging.atlassian.net/browse/VOGRE-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annotations/exceptions.py
```diff
@@ -30,4 +30,4 @@ def custom_403_handler(request, exception):
         'error_message': error_message
     }
 
-    return render(request, 'annotations/forbidden_error_page.html', context, status=403)
+    return render(request, 'annotations/forbidden_error_page.html', context, status=403)
```
jophals, line 33: did something here change?

annotations/templates/annotations/header.html
```diff
@@ -88,9 +88,7 @@
                         <ul class="dropdown-menu">
                             <li><a href="{% url "concepts" %}">Concepts</a></li>
                             <li><a href="{% url "relations" %}">Annotations</a></li>
-                            {% if user.is_staff %}
-                                <li><a href="{% url "list_relationtemplate" %}">Templates</a></li>
-                            {% endif %}
+                            {% if user.is_staff or user.is_vogon_admin %}<li><a href="{% url "list_relationtemplate" %}">Templates</a></li>{% endif %}
```
jophals: probably best to be consistent with your quoting syntax. Various {% %} blocks utilize double quotes inside and outside whereas some utilize double quotes outside and single quotes inside (I think single quotes inside is generally the popular choice)

annotations/templates/annotations/header.html
```diff
@@ -111,28 +109,37 @@
...
-                            </ul>
-                        {% else %}
-                            <a class="btn btn-sm" href="{% url 'account_login' %}" data-toggle="tooltip" title="to get started!" data-container="body" data-placement="bottom">
-                                <i class="fa fa-sign-in" aria-hidden="true"></i> {% trans 'Log in' %}
-                            </a>
-                        {% endif %}
-                    </li>
-                </ul>
-            </div><!--/.navbar-collapse -->
-        </div><!--/.container-fluid -->
```
jophals, line 135: what happened to this div (container-fluid)? Is it not needed? Because the opening div tag is still there but it's never closed

annotations/decorators.py
```diff
@@ -0,0 +1,29 @@
+from django.contrib.auth.decorators import login_required
+from django.core.exceptions import PermissionDenied
+from functools import wraps
+
+def vogon_admin_required(view_func):
```
jophals: is there a situation where a Django admin (staff) would be denied access but a Vogon admin granted access? Based on the Admin User management access toggling, it seems that Django admin trumps Vogon admin...

annotations/exceptions.py
Girik1105: Nothing changed here, I thought its a blank space but that didn't fix the issue when I added one.

jdamerow: It's probalby spaces vs tabs is my guess.

annotations/views/user_views.py
```diff
@@ -394,3 +402,40 @@ def list_user(request):
         'title': 'Contributors'
     }
     return render(request, template, context)
+
+@staff_member_required
```
jdamerow: I think this should be admin. I don't think we really use the staff role at this point, do we?

annotations/views/user_views.py
```diff
@@ -394,3 +402,40 @@ def list_user(request):
...
+    """
+    all_users = VogonUser.objects.all().order_by('username')
+    paginator = Paginator(all_users, settings.VOGON_ADMIN_PAGE_SIZE)
+
+    current_page_number = request.GET.get('page')
+    current_page = paginator.get_page(current_page_number)
+
+    return render(request, 'annotations/user_vogon_admin_list.html', {'page_obj': current_page})
+
+@staff_member_required
```
jdamerow: same here

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

PR #33, opened 2024-12-16
Story/VOGRE-32

rajvi-patel-22:

Workflow

User goes to concept list page

User selects a concept that has not be resolved

If there is a viaf url, Vogon will search Conceptpower for that viaf url and show the result. If there is exactly one result, user should be able to say “approve”, if there are more, user should select one

User should also have the chance to do a conceptpower search for a search term and select one of the results.

If there is no viaf url, user should get a  search  field to search conceptpower and then be able to select one of the results or do a new search

For all of these options, there should also be a button to create a new concept in conceptpower.

Once a concept has been selected and approved, Vogon concept is RESOLVED

We will likely only have “Pending” and “resolved” as concept states.

... Put ticket description here and add link to ticket ...
https://diging.atlassian.net/browse/VOGRE-32

Below bug is fixed along with this PR:
- Fix this.label for new concept name/label


concepts/lifecycle.py
```diff
@@ -261,38 +209,32 @@ def add(self):
...
+                "pos": pos,
+                "conceptlist": self.DEFAULT_LIST,
+                "description": self.instance.description,
+                "type": concept_type,
+                "equal_to": equal_uri
+            }
+            response = requests.post(url=url, data=json.dumps(concept_data), auth=auth)
+            
+            if response.status_code != requests.codes.ok:
+                raise RuntimeError(response.status_code, response.text)
```
jdamerow: this should happen in a conceptpower class not directly here.

concepts/lifecycle.py
```diff
@@ -230,10 +172,16 @@ def merge_with(self, uri):
         children_queryset.update(merged_with=target)
 
     def add(self):
+
```
jdamerow: remove empty line

concepts/lifecycle.py
```diff
@@ -304,16 +246,64 @@ def get_similar(self):
...
+            }
+            response = requests.get(url, headers=headers, params=parameters)
+            
+            if response.status_code == 200:
+                data = response.json()
+                concepts = []
+                if 'conceptEntries' in data:
+                    for concept_entry in data['conceptEntries']:
+                        concept = self.parse_concept(concept_entry)
+                        concepts.append(concept)
```
jdamerow: same as above, should not happen in this class

concepts/lifecycle.py
```diff
@@ -304,16 +246,64 @@ def get_similar(self):
...
+                'Accept': 'application/json',
+            }
+            response = requests.get(url, headers=headers, params=parameters)
+            if response.status_code == 200:
+                data = response.json()
+                concepts = []
+                if 'conceptEntries' in data:
+                    for concept_entry in data['conceptEntries']:
+                        concept = self.parse_concept(concept_entry)
+                        concepts.append(concept)
```
jdamerow: same

concepts/lifecycle.py
```diff
@@ -325,7 +315,43 @@ def get_matching(self):
...
+            headers = {
+                'Accept': 'application/json',
+            }
+            response = requests.get(url, headers=headers)
+
+            if response.status_code == 200:
+                    data = response.json()
+                    concept_entry = data.get('conceptEntries', [{}])[0]
+            else:
+                raise ValueError(f"Error fetching concept data: {response.status_code}")
```
jdamerow: same, there should be a conceptpower class that does all the communication with conceptpower (isn't there one already?)

jdamerow (changes requested):

concepts/conceptpower.py
```diff
@@ -0,0 +1,87 @@
...
+class Conceptpower:
+    # Default behavior is to leave these to the constructor.
+    endpoint = None
+    namespace = None
+
+    def __init__(self, **kwargs):
+        # Give first priority to the class definition, if endpoint or namespace are defined (and not None).
+        if self.endpoint is None:
+            self.endpoint = kwargs.get(
+                "endpoint", "http://chps.asu.edu/conceptpower/rest/")
```
jdamerow: these should not be hard-coded; if a default is needed it should come from environment variables/settings.

concepts/conceptpower.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if self.namespace is None:
+            self.namespace = kwargs.get(
+               "namespace", "{http://www.digitalhps.org/}")
```
jdamerow: same here

concepts/conceptpower.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def search(self, params=None, headers=None):
+        url = "{0}ConceptSearch".format(self.endpoint)
+
+        response = requests.get(url, headers=headers, params=params)
+        concepts = []
+        if response.status_code == 200:
```
jdamerow: if not, we definitely want to log what else is returned and an error message if there is one.

concepts/conceptpower.py
```diff
@@ -0,0 +1,87 @@
...
+                    for concept_entry in data['conceptEntries']:
+                        concept = self.parse_concept(concept_entry)
+                        concepts.append(concept)
+        return concepts
+
+    def get(self, uri, headers):
+        url = "{0}Concept?id={1}".format(self.endpoint, uri)
+        response = requests.get(url, headers=headers)
+        data = {}
+        if response.status_code == 200:
```
jdamerow: same here

concepts/conceptpower.py
jdamerow: also, you want to use a constant (`requests` probably has one) and not a magic number (even though it's an obvious one if you have done web application development before).

concepts/conceptpower.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if r.status_code != requests.codes.ok:
+            raise RuntimeError(r.status_code, r.text)
+
+        # Returned data after successful response
+        return r.json()
+    
+    def parse_concept(self,concept_entry):
+        """
+        Parse a concept and return a dictionary with the required fields.
```
jdamerow, line 79: needs parameter description (what does this method accept to get in `concept_entry`? maybe even put in example json (think about what would someone need to know to use this method if they don't see the code)).

jdamerow (changes requested):

concepts/conceptpower.py
```diff
@@ -0,0 +1,132 @@
+import requests,json
+import xml.etree.ElementTree as ET
+from requests.auth import HTTPBasicAuth
+
+
+class Conceptpower:
+    # Default behavior is to leave these to the constructor.
```
jdamerow: what does this mean?

concepts/conceptpower.py
```diff
@@ -0,0 +1,132 @@
...
+    endpoint = None
+    namespace = None
+
+    def __init__(self, **kwargs):
+        # Give first priority to the class definition, if endpoint or namespace are defined (and not None).
```
jdamerow: what does that mean?

vogon/settings.py
```diff
@@ -198,8 +198,8 @@
 
 CONCEPTPOWER_USERID = os.environ.get('CONCEPTPOWER_USERID')
 CONCEPTPOWER_PASSWORD = os.environ.get('CONCEPTPOWER_PASSWORD')
-CONCEPTPOWER_ENDPOINT = os.environ.get('CONCEPTPOWER_ENDPOINT')
-CONCEPTPOWER_NAMESPACE = os.environ.get('CONCEPTPOWER_NAMESPACE')
+CONCEPTPOWER_ENDPOINT = os.environ.get('CONCEPTPOWER_ENDPOINT', 'http://chps.asu.edu/conceptpower/rest/')
+CONCEPTPOWER_NAMESPACE = os.environ.get('CONCEPTPOWER_NAMESPACE', '{http://www.digitalhps.org/}')
```
jdamerow: why is the namespace in curly brackets?

vogon/settings.py
```diff
@@ -198,8 +198,8 @@
 
 CONCEPTPOWER_USERID = os.environ.get('CONCEPTPOWER_USERID')
 CONCEPTPOWER_PASSWORD = os.environ.get('CONCEPTPOWER_PASSWORD')
-CONCEPTPOWER_ENDPOINT = os.environ.get('CONCEPTPOWER_ENDPOINT')
-CONCEPTPOWER_NAMESPACE = os.environ.get('CONCEPTPOWER_NAMESPACE')
+CONCEPTPOWER_ENDPOINT = os.environ.get('CONCEPTPOWER_ENDPOINT', 'http://chps.asu.edu/conceptpower/rest/')
```
jdamerow: and let's not put that url here. even though it is in the github commit history now, we want to keep bot traffic to our servers as low as possible

vogon/settings.py
jdamerow: this one is fine, it's just a static github page

jdamerow (changes requested):

vogon/settings.py
rajvi-patel-22: > why is the namespace in curly brackets?

Based on the old code and the places where it has been used, it seems to be enclosed in curly braces..
I have referred `conceptpower-api` repo ([here](https://github.com/diging/Conceptpower-API/blob/master/conceptpower/classes.py#L36))

Places where it's being used:  [here](https://github.com/diging/vogon-resuscitate/blob/develop/concepts/authorities.py#L131)

jdamerow: That doesn't seem to make sense. The `get_namespace` puts curly brackets around the namespace when returning, is that needed anywhere? why not drop them altogether (in the method and the environment variable)? It's not logical at all why I would add curly brackets around a namespace when configuring the application.

jdamerow:
Make it so, Jenkins.

jdamerow:
this seems to break Vogon:
```
File "<frozen importlib._bootstrap_external>", line 728, in exec_module
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  |   File "/vogon/concepts/admin.py", line 10, in <module>
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  |     from . import authorities
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  |   File "/vogon/concepts/authorities.py", line 6, in <module>
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  |     from conceptpower import Conceptpower
Feb 06 15:14:56 ubuntu20 docker[259738]: web-1  | ModuleNotFoundError: No module named 'conceptpower'
```

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
When I try to resolve a concept, it asks me to login as admin. That shouldn't happen if I'm a Vogon admin.
Also, in the filter dropdown for concepts, do all these statuses still exist (e.g. approved)?

jdamerow:
Make it so, Jenkins.

jdamerow:
I'm getting a 401 when I try to resolve a concept. My user is a Vogon Admin.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow (commented):
After a concept has been merge, the uri that is shown should be the Concetppower one, but I'm going to merge this and create a new ticket for this.

Issue #46, opened 2025-01-07
Unable to edit/delete appellations once created in Text Annotation View


Warrior1412:
There is no way to edit the created notation. One a notation is created, it is not addressed to edit / how to edit the created notation.

PR #54, opened 2025-01-31
story/VOGRE-54

Girik1105:

Unable to edit/delete appellations once created in Text Annotation View

There is no way to edit the created notation. One a notation is created, it is not addressed to edit / how to edit the created notation.

https://github.com/diging/vogon-resuscitate/issues/46

[VOGRE-54](https://diging.atlassian.net/browse/VOGRE-54?atlOrigin=eyJpIjoiYmU1MmFhYmZmOTYzNGU3OWJiMzU2MzZhMTBkYjUxMjAiLCJwIjoiaiJ9)

No
  
... describe here ...

[VOGRE-54]: https://diging.atlassian.net/browse/VOGRE-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annotations/views/rest_views.py
```diff
@@ -321,6 +333,84 @@ def create(self, request, *args, **kwargs):
...
+            if 'stringRep' in request.data:
+                instance.stringRep = request.data['stringRep']
+            
+            # Update interpretation if provided
+            if 'interpretation' in request.data:
+                try:
+                    concept = Concept.objects.get(uri=request.data['interpretation'])
+                    instance.interpretation = concept
+                except Concept.DoesNotExist:
+                    pass
```
jdamerow: an appellation without a concept should not be savable.

annotations/views/rest_views.py
```diff
@@ -321,6 +333,84 @@ def create(self, request, *args, **kwargs):
...
+                          object_object_id=instance.id) |
+                        Q(predicate=instance)
+                    )
+                ).exists()
+            )
+
+            if in_relations:
+                return Response(
+                    {"detail": "Cannot delete an Appellation used in a Relation."},
+                    status=status.HTTP_403_FORBIDDEN
```
jdamerow: Forbidden is used for permission issues, here we maybe want to use 409 Conflict ("The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request. ")

annotations/views/rest_views.py
```diff
@@ -321,6 +333,84 @@ def create(self, request, *args, **kwargs):
...
+                )
+            
+            self.perform_destroy(instance)
+            return Response(status=status.HTTP_204_NO_CONTENT)
```
jdamerow: if this indicates that the appellation has been deleted, it should return OK 200.

annotations/views/rest_views.py
```diff
@@ -321,6 +333,84 @@ def create(self, request, *args, **kwargs):
...
+        
+        except Exception as e:
+            logger.error(f"Error deleting appellation: {str(e)}", exc_info=True)
+            return Response(
+                {"detail": f"Error deleting appellation: {str(e)}"},
+                status=status.HTTP_400_BAD_REQUEST
```
jdamerow: if this is for any exception that may happen, it should be 500.

jdamerow (changes requested):

annotations/views/rest_views.py
```diff
@@ -321,6 +333,93 @@ def create(self, request, *args, **kwargs):
...
+            instance.save()
+            
+            serializer = self.get_serializer(instance)
+            return Response(serializer.data)
+            
+        except Exception as e:
+            logger.error(f"Error updating appellation: {str(e)}", exc_info=True)
+            return Response(
+                {"error": str(e)},
+                status=status.HTTP_400_BAD_REQUEST
```
jdamerow, line 392: 400 should only be sent when the client could change something to make this request go through, but here you are catching any exception, so it could an exception that has nothing to do with what the client sends? either this should return 500 or different exceptions should be handled differently.

annotations/views/rest_views.py
```diff
@@ -321,6 +333,93 @@ def create(self, request, *args, **kwargs):
...
+                        Q(object_content_type=ContentType.objects.get_for_model(Appellation),
+                          object_object_id=instance.id) |
+                        Q(predicate=instance)
+                    )
+                ).exists()
+            )
+            if in_relations:
+                return Response(
+                    {"detail": "Cannot delete an Appellation used in a Relation."},
+                    status=status.HTTP_409_CONFLICT
```
jdamerow: conflicts are typically send with there is a version of the sent resource that is different than the current one in the system (or different from what the sender based their changes on), so more like a GitHub conflict. Here I would send a bad request (since this should not be allowed in the first place to delete appellations used in a relation).

jdamerow (changes requested):

jdamerow:
Added Julian as reviewer especially for the Vue stuff.

annotations/static/annotations/js/annotators/appellationdisplay.js
```diff
@@ -60,20 +65,71 @@ AppellationDisplayItem = {
...
+                    this.$el.parentNode.removeChild(this.$el);
+                }
+                
+                // Clear all stores to ensure clean state
+                store.commit('removeAppellation', deletedAppellation);
+                store.commit('setTextAppellation', []);
+                store.commit('resetCreateAppelltionsToText');
+                
+                // Notify parent components about deselection
+                this.$root.$emit('appellationDeselected', this.appellation);
```
jophals: It's fairly conflicting to use both Vuex and the `$root` instance to manage state. Generally considered a better idea to use Vuex but I would stick to one.

jdamerow:
and now also conflicts

jophals:
i still see a lot of $root usage instead of vuex

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -304,5 +375,32 @@ AppellationList = {
...
+        },
+        /*
+         * Updates an appellation in the current_appellations array
+         * @param updatedAppellation - The new appellation data to update with
+         * 
+         * This function:
+         * 1. Finds the appellation to update by matching IDs
+         * 2. Creates a new array to maintain reactivity
+         * 3. Replaces the old appellation with the updated one
+         * 4. Assigns the new array back to current_appellations
```
jdamerow, line 456: Am I missing something, or this this not done? As far as I can tell, only 2. and 3. are done?

annotations/static/annotations/js/annotators/textdisplay.js
```diff
@@ -194,31 +242,35 @@ TextDisplay = {
...
-                    //  may have been left-to-right or right-to-left.
+                    // Get the start and end position of the selection
                     var selection = document.getSelection();
                     
                     // Make sure we have a valid selection
                     if (!selection || selection.rangeCount === 0) return;
                     
                     var startOffset = Math.min(selection.anchorOffset, selection.focusOffset);
                     var endOffset = Math.max(selection.anchorOffset, selection.focusOffset);
+                    console.log('Selection offsets:', startOffset, endOffset);
```
jdamerow: is this needed?

annotations/static/annotations/js/annotators/textdisplay.js
```diff
@@ -227,49 +279,141 @@ TextDisplay = {
...
+                                EventBus.$emit('showMessage', {
+                                    text: 'Successfully updated annotation',
+                                    type: 'success'
+                                });
+                            }).catch(error => {
+                                console.error('Failed to update appellation:', error);
+                                EventBus.$emit('showMessage', {
+                                    text: 'Failed to update annotation. Please try again.',
+                                    type: 'error'
+                                });
```
jdamerow, line 355: shouldn't editing be canceled in this case as well (especially if the mouse selection is cleared afterwards)?

jdamerow (changes requested):

annotations/static/annotations/js/annotators/appellationlist.js
Girik1105: `const newAppellations = [...this.current_appellations];`

This line uses the spread operator(...) to create new array that contains all the elements from current_appellations. Vue's reactivity system works by tracking the changes to data and might not trigger changes such as edit highlights.

jdamerow: Yes, I understand that it creates a new array, but it never seems to assign the new array back to current_appellations?

Girik1105: Apologies for the oversight, I just added a line to reassign current appellations with newAppellations

annotations/static/annotations/js/annotators/textdisplay.js
```diff
@@ -227,49 +278,144 @@ TextDisplay = {
...
+                            });
+                            
+                            // Clear selection after update attempt
+                            clearMouseTextSelection();
+                            return; // Important: don't proceed to normal selection handling
+                        }
+                    }
+
+                    // Normal text selection handling (only if not in edit mode)
+                    if (!self.isEditing) {
```
jdamerow, line 379: isn't this always executed? Either isEditing is already false, or isEditing will be set to false after an update?

annotations/static/annotations/js/annotators/textdisplay.js
```diff
@@ -227,49 +278,144 @@ TextDisplay = {
...
+                                self.resetTextSelection();
+                                
+                                // Show success message
+                                EventBus.$emit('showMessage', {
+                                    text: 'Successfully updated annotation',
+                                    type: 'success'
+                                });
+                            }).catch(error => {
+                                console.error('Failed to update appellation:', error);
+                                EventBus.$emit('cancelEdit');
```
jdamerow, line 351: if we cancel edit here, should it also do:
```
 self.isEditing = false;
```

annotations/views/rest_views.py
```diff
@@ -44,16 +45,19 @@ def has_permission(self, request, view):
...
-        # Allow GET requests for authenticated users
         if request.method in ['GET', 'HEAD', 'OPTIONS']:
             return True
 
-        # check if user is owner or collaborator For POST/PUT/DELETE 
+        if request.method == 'DELETE':
+            return True  # Let has_object_permission handle it
+
+        if request.method in ['PUT', 'PATCH']:
+            return True  # Let has_object_permission handle it
```
jdamerow: why not add all of the above to line 48? with a comment that includes the comments here?

jdamerow (changes requested):

annotations/static/annotations/js/annotators/textdisplay.js
Girik1105: Since we emit cancelEdit here, our EventBus.$on('cancelEdit') listener already does that.

```
        EventBus.$on('cancelEdit', () => {
            this.isEditing = false;
            localStorage.removeItem('editingAppellation');
            document.removeEventListener('keydown', this.handleEscKey);
            // Clear any existing selection
            this.resetTextSelection();
        });

```

annotations/static/annotations/js/annotators/textdisplay.js
Girik1105: this condition is not always executed. When the code is in edit mode and processing an appellation update, there's an early return statement that prevents execution from reaching this point.

This check also performs to major checks

1.  When NOT in edit mode (isEditing is false): The condition passes, and the code emits the 'selecttext' event, which initiates the process of creating a new appellation with the selected text.

2. When IN edit mode (isEditing is true): The condition fails, preventing the 'selecttext' event from firing, which ensures you don't accidentally create a new appellation while trying to update an existing one.

annotations/static/annotations/js/annotators/textdisplay.js
jdamerow: but on line 361-362 you do both. should be consistent

annotations/static/annotations/js/annotators/textdisplay.js
jdamerow: ah, I see now that the return only returns from the api call code, not the whole block.

annotations/static/annotations/js/annotators/appellationdisplay.js
```diff
@@ -77,6 +77,22 @@ AppellationDisplayItem = {
...
+                // Update the appellation data
+                Object.assign(this.appellation, updatedAppellation);
+                
+                // Ensure the appellation is visible
+                this.appellation.visible = true;
+                
+                // Force a refresh of positions for updated appellation
+                this.$nextTick(() => {
+                    this.updatePosition();
+                    EventBus.$emit('updateposition');
```
jdamerow, line 92: where is this emitted event caught? Is that here?
https://github.com/diging/vogon-resuscitate/blob/935a07582e8b9bf165683842c613de3408feded9/annotations/static/annotations/js/annotators/textdisplay.js#L46
If so, wouldn't there be two `updatePosition` methods be called that are very similar?

https://github.com/diging/vogon-resuscitate/blob/935a07582e8b9bf165683842c613de3408feded9/annotations/static/annotations/js/annotators/textdisplay.js#L65

https://github.com/diging/vogon-resuscitate/blob/935a07582e8b9bf165683842c613de3408feded9/annotations/static/annotations/js/annotators/appellationdisplay.js#L114

Also, why are there two methods that are almost but not quite identical?

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -119,8 +154,44 @@ var AppellationListItem = {
...
+                // Emit event to parent to remove from list and text display
+                this.$emit('removeappellation', this.appellation);
+            }).catch(error => {
+                if (error.status === 400) {
+                    this.deleteError = "This annotation is used in a relation and cannot be deleted.";
+                } else {
+                    this.deleteError = "Error deleting annotation, Please try again later.";
+                }
+                
+                // Clear error message after 5 seconds
```
jdamerow: Error messages should not be automatically cleared. What if the user turns around or goes to get coffee and misses the error message?

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -168,8 +239,8 @@ AppellationList = {
 									v-on:hideappellation="hideAppellation"
 									v-on:showappellation="showAppellation"
 									v-on:selectappellation="selectAppellation"
-									v-on:removeAppellation="removeAppellation($event)"
-									v-on:addAppellation="addAppellation($event)"
+                                    v-on:addAppellation="addAppellation($event)"
```
jdamerow, line 302: indentation?

jdamerow (changes requested):

annotations/static/annotations/js/annotators/appellationdisplay.js
jdamerow: This still. There seem to be two functions that are very similar.

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -1,129 +1,265 @@
...
-            } else {
+            } else if (creator && creator.username) {
                 return creator.username;
             }
+            return 'Unknown';
         },
         getFormattedDate: function (isodate) {
-            return moment(isodate).format('dddd LL [at] LT');
-        }
+            if (typeof moment !== 'undefined') {
```
jdamerow: when would this happen? wouldn't this be something that can be avoided by ensuring `moment` is included?

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -1,129 +1,265 @@
...
+        },
 
-    }
-}
+        deleteAppellation: function() {
+            this.deleteError = null;
+            const deletedAppId = this.appellation.id; 
+
+            if (typeof Appellation === 'undefined' || !Appellation.delete) {
+                this.deleteError = "Cannot delete: Appellation resource unavailable.";
```
jdamerow: What does this mean? Does this mean the appellation doesn't exist on the server? This is the error message that's being shown to the user, right? So it should be something the user can understand.

annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -1,129 +1,265 @@
...
+            Appellation.delete({id: deletedAppId}).then(response => {
+                this.$emit('removeappellation', this.appellation); 
+
+                const editingAppData = localStorage.getItem('editingAppellation');
+                let isDeletingTheCurrentlyEditedItem = false;
+                if (editingAppData) {
+                    try {
+                        const editingApp = JSON.parse(editingAppData);
+                        // Check if the parsed data has an id and it matches the deleted item's id
+                        if (editingApp && typeof editingApp.id !== 'undefined' && editingApp.id === deletedAppId) {
```
jdamerow, line 159: is there a reason to use `typeof editingApp.id !== 'undefined'` over `editingApp.id !== undefined`?

annotations/static/annotations/js/annotators/resources.js
jdamerow, line 1: is there anything in this file that was actually changed? if no, it should not be changed.

jdamerow (changes requested):

annotations/static/annotations/js/annotators/appellationlist.js
Girik1105: the typeof function returns the string "undefined" for variables that are declared but not initialized, or for object properties that don't exist so using ` "undefined"` just adds an extra layer of checking.

annotations/static/annotations/js/annotators/appellationdisplay.js
Girik1105: The functions do perform similar tasks, they take a representation text selection and use the utility function `calculateOverlayPositions` to convert those positions into screen coordinates suitable for showing the highlight or overlay over the text.

But they have a few differences:

updatePosition in AppellationDisplay calculates overlays for pre-existing, saved appellations (using this.appellation.position).

TextDisplay.js calculates overlays for the current, user selection (using this.selected) typically for creating new annotations or providing real-time selection feedback. It also includes logic to clear the selection display.

annotations/static/annotations/js/annotators/resources.js
Girik1105: Im not sure why this is happening, I didn't change anything in this file. I even tried to copy paste the file from develop, but it still shows this.

annotations/static/annotations/js/annotators/appellationdisplay.js
```diff
@@ -96,42 +118,15 @@ AppellationDisplayItem = {
...
-                this.end_position = {};
+            if (!this.appellation || !this.appellation.position) {
+                return;
             }
+
+            var calc = calculateOverlayPositions(this.appellation.position);
+            this.position     = calc.position;
+            this.mid_lines    = calc.mid_lines;
+            this.end_position = calc.end_position;
+            this.line_height  = calc.line_height;
```
jdamerow: this exact same code is also called in textdisplay.js:
```
this.position     = calc.position;
this.mid_lines    = calc.mid_lines;
this.end_position = calc.end_position;
this.line_height  = calc.line_height;
```

annotations/static/annotations/js/annotators/appellationdisplay.js
jdamerow: I'm still confused. This code:
```
this.updatePosition();
EventBus.$emit('updateposition');
```
first calls below updatePosition function, right? Then it emits an event `updateposition`. Won't this event be caught in textdisplay.js line 45, which will then call the updatePosition function in textdisplay.js?

Girik1105:
So out here each component only updates its own position data. They don't call each other's methods.

What both the updatePosition functions do:

1. TextSelectionDisplay.updatePosition() (textdisplay.js lines 67-78)
Creates temporary grey/blue highlight overlays when user selects text to create new appellations. When a user is selecting the text for updating an existing appellation.

2. AppellationDisplayItem.updatePosition() (appellationdisplay.js lines 119-125):
Creates permanent colored highlight overlays for existing saved appellations. This pulls the position from the database, this is the yellow highlight of the saved appellations

Workflow for when a user creates a new appellation:

INITIAL STATE:
- User sees text and any existing colored highlights (permanent appellations)
- No current text selection (no blue highlight)

1. User drags mouse to select text (e.g., offsets 100–200)
   ↓
2. TextDisplay.handleMouseup() captures the selection:
   - Sets: this.selected = { startOffset: 100, endOffset: 200, representation: "Ferdinand Schmutzer" }
   ↓
3. TextSelectionDisplay.updatePosition() triggers automatically (because it watches 'selected'):
   - Converts selection offsets (100–200) to screen coordinates
   - Updates its own data to show a blue.grey highlight over the selected text
   ↓
4. User confirms creation
   - Appellation is saved to the database with position 100–200
   ↓
5. New AppellationDisplayItem is created for the new appellation:
   - Calls AppellationDisplayItem.updatePosition()
   - Shows a permanent colored highlight at 100–200
   ↓
6. EventBus.$emit('updateposition') is fired
   - Notifies all components (TextSelectionDisplay, all AppellationDisplayItems) to recalculate their positions
   - Ensures all highlights are perfectly synchronized
   ↓
7. TextSelectionDisplay.selected is cleared (no selection)
   - Blue highlight disappears

FINAL STATE:
- The new colored highlight (permanent) is visible at 100–200
- No blue highlight (no active selection)
- All highlights are in sync with the current layout



The Workflow for editing an appellation and how each function works:

INITIAL STATE:
- Appellation 1 highlighted at positions 150-200 example: "Ferdinand Schmutzer"
- No current text selection

1. User clicks the edit icon on Appellation 1
   ↓
2. EditAppellation() function:
   - localStorage.setItem('editingAppellation', JSON.stringify(appellation))
   - EventBus.$emit('startEdit') → Sets isEditing = true
   ↓
3. User selects NEW text at positions 300-350 ("Albert Einstein")
   ↓
4. TextDisplay.handleMouseup() in edit mode:
   - Creates selection: { startOffset: 300, endOffset: 350, representation: "Albert Einstein" }
   ↓
5. TextSelectionDisplay.updatePosition() triggers (watch: selected):
   - calculateOverlayPositions({ startOffset: 300, endOffset: 350 })
   - applyOverlayPositions(this, calc)
   - Shows BLUE highlight at positions 300-350 (temporary)
   ↓
6. Server update happens:
   - Appellation.update() sends new position_value: "300,350"
   - Backend responds with updated appellation data
   ↓
7. AppellationDisplayItem receives 'appellationUpdated' event:
   - Object.assign(this.appellation, updatedAppellation)
   - this.appellation.position.startOffset = 300
   - this.appellation.position.endOffset = 350
   ↓
8. AppellationDisplayItem.updatePosition() fires:
   - calculateOverlayPositions(this.appellation.position) // 300-350
   - applyOverlayPositions(this, calc)
   - Moves COLORED highlight from 150-200 → 300-350
   - EventBus.$emit('updateposition') // Notify others
   ↓
9. TextSelectionDisplay receives 'updateposition' event:
   - sleep(200).then(self.updatePosition)
   - TextSelectionDisplay.updatePosition() fires:
     - Recalculates current selection highlight (if any)
     - Ensures blue highlight stays accurate
   ↓
10. Edit cleanup:
    - localStorage.removeItem('editingAppellation')
    - EventBus.$emit('cancelEdit') → Sets isEditing = false
    - Blue highlight disappears (selection cleared)

FINAL STATE:
- Appellation 1 is now highlighted from positions 150-200 ("Ferdinand Schmutzer") to positions 300-350 ("Albert Einstein")
- No current text selection

Why TextSelectionDisplay Needs to Update:
- When an appellation changes position, it can affect the text layout in ways that make the current text selection highlight appear in the wrong place. 

Different updatePosition() Methods
- They're different functions on different component instances:
- AppellationDisplayItem.updatePosition() - Updates permanent appellation highlights
- TextSelectionDisplay.updatePosition() - Updates temporary text selection highlights

Its not a circular loop:
- TextSelectionDisplay.updatePosition() (line 62-73) does NOT emit any events
- It only updates its own position data and stops
- The chain ends there

PR #60, opened 2025-03-04
story/VOGRE-38

rajvi-patel-22:

During concept resolution, a user should be able to flag a concept that would then show up as needing resolution in a another page for the user who created it.
https://diging.atlassian.net/browse/VOGRE-38

  #33 

... describe here ...


concepts/views.py
```diff
@@ -59,6 +59,17 @@ def merge_concepts(request, source_concept_id):
...
+@login_required
+def flag_concept(request, source_concept_id):
+    concept = get_object_or_404(Concept, pk=source_concept_id)
+    if request.method == "POST":
+        comment = request.POST.get("flag_comment", "").strip()
+        concept.concept_state = Concept.FLAGGED    
+        concept.comment = comment
+        concept.save()
+        next_page = request.GET.get('next', reverse('concepts'))
+    return HttpResponseRedirect(next_page)
```
jdamerow: if request is not `POST` this will throw an exception since `next_page` does not exist.

concepts/views.py
```diff
@@ -127,7 +138,7 @@ def add_concept(request, concept_id):
         'next_page': urllib.parse.quote_plus(next_page),
         'back_to_page': back_to_page
     }
-    if concept.concept_state != Concept.PENDING:
+    if concept.concept_state != Concept.PENDING and concept.concept_state != Concept.FLAGGED:
```
jdamerow: This might be right but it looks wrong (I honestly can't tell at the moment). The `and` with the not equal is really throwing me off. What's the logic here? Only pending concepts should be added, right? Flagged concepts should not be added either, shouldn't it? So is this change needed?

jdamerow (changes requested):

concepts/views.py
rajvi-patel-22: This view function is triggered when a user clicks the `Resolve` button from the concept list page. It does not immediately add the concept to ConceptPower. Instead, it initiates the resolution workflow by checking for similar existing concepts. If similar concepts are found, the user is presented with the option to merge. If no suitable match exists, the user can proceed to either create a new concept or flag the current one.

The `if condition` ensures that the resolution process is only available for concepts that are still needs to be resolve. Concepts that are already reviewed or finalized (i.e., in a resolved state like RESOLVED or REJECTED) should not be reprocessed.

Maybe I should rewrite it for better readability using `not in` and also add a comment:
```
# Skip concepts that are already resolved (e.g., resolved or rejected).
if concept.concept_state not in [Concept.PENDING, Concept.FLAGGED]:
```

jdamerow: The condition should probably be turned around: `if xxx is true, then do something` not `if yyy is not true, then do something`.

concepts/views.py
```diff
@@ -127,30 +141,31 @@ def add_concept(request, concept_id):
...
-        'candidates': candidates,
-        'matches': matches,
-    })
-
-    return render(request, "annotations/concept_add.html", context)
+    # Process only if the concept is still in a resolvable state (e.g., PENDING or FLAGGED)
+    if concept.concept_state in [Concept.PENDING, Concept.FLAGGED]:
+        if request.GET.get('confirmed', False):
+            try:
+                manager.add()
```
jdamerow: `manager` is not a great name. I would expect that if this were truly a manager, then the add method would get the concept to add. I know this was already there but the boy scout rule applies: "leave the code in a better state than you found it in".

concepts/views.py
```diff
@@ -127,30 +141,31 @@ def add_concept(request, concept_id):
...
-
-    context.update({
-        'candidates': candidates,
-        'matches': matches,
-    })
-
-    return render(request, "annotations/concept_add.html", context)
+    # Process only if the concept is still in a resolvable state (e.g., PENDING or FLAGGED)
+    if concept.concept_state in [Concept.PENDING, Concept.FLAGGED]:
+        if request.GET.get('confirmed', False):
```
jdamerow: a GET request should not modify the state. creating a new concept should happen via a POST request.

concepts/models.py
```diff
@@ -48,16 +48,19 @@ class Concept(HeritableObject):
...
         (REJECTED, 'Rejected'),
         (RESOLVED, 'Resolved'),
         (MERGED, 'Merged'),
+        (FLAGGED, 'Flagged'),
     )
     concept_state=models.CharField(max_length=10, choices=concept_state_choices,
                                    default='Pending')
     merged_with = models.ForeignKey('Concept', related_name='merged_concepts',
                                     **optional, on_delete=models.CASCADE)
+    comment = models.TextField(blank=True, null=True)
```
jdamerow: This seems to narrow, this should probably a one to many relationship, so the user can leave more than one comment. At this point, it's fine to have only create a comment when flagging a concept, but I'll create a ticket so that other users can respond to comments.

jdamerow (changes requested):

concepts/models.py
jdamerow: What I meant is that the variable here should be able to accommodate multiple comments even though only one needs to be created for this ticket.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

PR #64, opened 2025-03-12
story/VOGRE-61

Girik1105:

Write tests for Vogon

[VOGRE-61](https://diging.atlassian.net/browse/VOGRE-61?atlOrigin=eyJpIjoiZjljZjk0ZTQ2NzZhNDE4YmE2NTgzZTFjY2FiODYzNGQiLCJwIjoiaiJ9)

... describe here ...

[VOGRE-61]: https://diging.atlassian.net/browse/VOGRE-61?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jdamerow:
The tests fail because the actions can only run Python 3.9 but not all dependencies work with Python 3.9 (e.g. anyjson, which apparently is abandoned. There is a PR to upgrade to 3.13, maybe here, let's go with 3.9 and then go from there to 3.13 in the other ticket.

jdamerow:
I think the workflow file needs to be updated to run with the correct python version.

jdamerow:
Tests are failing.

jdamerow:
Tests are failing.

Girik1105:
> Tests are failing.

All tests work now! There was an issue in the django.yml configuration and the POSTGRES configuration in settings.

PR #68, opened 2025-03-18
story/VOGRE-64

Girik1105:

On the collections page, when importing a text, collections need to be sorted alphabetically and should show how many items they contain

[VOGRE-64](https://diging.atlassian.net/browse/VOGRE-64?atlOrigin=eyJpIjoiOWQ5NmVlMzQ2YjNkNGRlNzhhOTkwYzU5MTI5MzZkNjAiLCJwIjoiaiJ9)

... describe here ...

[VOGRE-64]: https://diging.atlassian.net/browse/VOGRE-64?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/annotations/utils.py
+++ b/annotations/utils.py
@@ -176,3 +176,19 @@ def _annotate_project_counts(queryset):
                           distinct=True),
         num_collaborators=Count('collaborators', distinct=True)
     )
+
+
+"""
+Helper function to sort collections alphabetically by name and integer parts (case-insensitive)
+"""
+def natural_keys(text):
+    """
+    Converts a string into a list of strings and integers, so that it sorts numerically on the numeric parts.
+    Example: Turn 'test10' into ['test', 10] so it sorts numerically on the numeric parts.
+    """
+    # This inner function tries to convert digit-only substrings to integers.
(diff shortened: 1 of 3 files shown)


annotations/utils.py
```diff
@@ -175,4 +175,20 @@ def _annotate_project_counts(queryset):
...
         num_collaborators=Count('collaborators', distinct=True)
-    )
\ No newline at end of file
+    )
+
+
+"""
+Helper function to sort collections alphabetically by name and integer parts (case-insensitive)
+"""
+def natural_keys(text):
```
jdamerow, line 184: while I appreciate your thoughtfulness I don't think this is necessary, I doubt that the case we have in with test collections is very common. Let's keep this function since it might be coming in useful at some point, but let's sort collections just alphabetically.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

PR #36, opened 2025-01-03
story/VOGRE-36

rajvi-patel-22:
# Guidelines for Pull Requests

If you haven't yet read our code review guidelines, please do so, You can find them [here](https://diging.atlassian.net/wiki/spaces/DIGING/pages/2256076801/Code+Review+Guidelines).

Please confirm the following by adding an x for each item (turn `[ ]` into `[x]`).

- [x] I have removed all code style changes that are not necessary (e.g. changing blanks across the whole file that don’t need to be changed, adding empty lines in parts other than your own code)
- [x] I am not making any changes to files that don’t have any effect (e.g. imports added that don’t need to be added)
- [x] I do not have any sysout statements in my code or commented out code that isn’t needed anymore
- [x] I am not reformatting any files in the wrong format or without cause. 
- [x] I am not changing file encoding or line endings to something else than UTF-8, LF
- [x] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed
- [x] I have added Javadoc/documentation where appropriate
- [x] I have added test cases where appropriate
- [x] I have explained any part of my code/implementation decisions that is not be self-explanatory

## Please provide a brief description of your ticket 
Right now only admins see the concept list and then all concepts for all users, but users should be able to approve their own concept.
https://diging.atlassian.net/browse/VOGRE-36

## Are there any other pull requests that this one depends on?
https://github.com/diging/vogon-resuscitate/pull/33
  
## Anything else the reviewer needs to know?

... describe here ...

PR #71, opened 2025-03-25
story/VOGRE-37

rajvi-patel-22:

Users should set up their own connection with Conceptpower by entering their username and password and during concept resolution, their own account would be used. 

https://diging.atlassian.net/browse/VOGRE-37

  This PR depends on https://github.com/diging/vogon-resuscitate/pull/36

... describe here ...


concepts/lifecycle.py
```diff
@@ -206,12 +205,14 @@ def add(self):
...
         try:
-            data = self.conceptpower.create(self.user, self.password,
+            data = self.conceptpower.create(self.user,
                                             self.instance.label, pos,
                                             self.DEFAULT_LIST,
                                             self.instance.description,
                                             concept_type,
                                             equal_to=equal_uri)
+        except ConceptPowerCredentialsMissingException as e:
+            raise e
```
jdamerow: since this exception is already thrown I don't think you need to rethrow it?

external_accounts/views.py
```diff
@@ -157,4 +159,50 @@ def citesphere_disconnect(request, repository_id):
...
+    if request.method == 'POST':
+        new_password = request.POST.get("new_password")
+
+        try:
+            account = ConceptpowerAccount.objects.get(user=request.user)
+            account.password = new_password
+            account.save()
+            messages.success(request, "Password updated successfully!")
+        except ConceptpowerAccount.DoesNotExist:
+            messages.error(request, "No ConceptPower account found.")
```
jdamerow: conceptpower is spelled "Conceptpower"

external_accounts/views.py
```diff
@@ -157,4 +159,50 @@ def citesphere_disconnect(request, repository_id):
...
+
+    return redirect(reverse('dashboard'))
+
+@login_required
+def conceptpower_disconnect(request):
+    try:
+        ConceptpowerAccount.objects.filter(user=request.user).delete()
+    except Exception as e:
+        print(f"Error disconnecting Conceptpower account: {e}")
+
```
jdamerow, line 208: the user should see some message saying that either the account was disconnected successfully or that it didn't work.

jdamerow (changes requested):

concepts/conceptpower.py
```diff
@@ -1,7 +1,10 @@
 import requests,json
 from requests.auth import HTTPBasicAuth
+from external_accounts.models import ConceptpowerAccount
 
-
+class ConceptPowerCredentialsMissingException(Exception):
```
jdamerow: Conceptpower is misspelled (lowercase 'p')

concepts/conceptpower.py
```diff
@@ -49,10 +52,18 @@ def get(self, uri, headers):
             raise ValueError(f"Error fetching concept data: {response.status_code}")
         return concept_entries[0] if concept_entries else {}
 
-    def create(self, user, password, label, pos, conceptlist, description,
+    def create(self, user, label, pos, conceptlist, description,
```
jdamerow, line 55: just as important as throwing an exception when conceptpower credentials are missing is to throw one when the authentication is incorrect. right now it looks like it would just show a generic 'something went wrong' message to the user.

external_accounts/views.py
```diff
@@ -157,4 +159,51 @@ def citesphere_disconnect(request, repository_id):
...
+        password = request.POST.get('password')
+        next_url = request.POST.get('next', reverse('dashboard'))
+
+        if not username or not password:
+            messages.error(request, "Both username and password are required.")
+            return render(request, 'login/login.html')
+
+        account, created = ConceptpowerAccount.objects.get_or_create(user=request.user)
+        account.username = username
+        account.password = password
```
jdamerow, line 177: password should not be stored in plaintext. it should be encrypted in a way that we can get the original password back, but at least that.

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

external_accounts/models.py
```diff
@@ -54,3 +56,36 @@ def fetch_citesphere_user_id(sender, instance, created, **kwargs):
...
+    _encryption_key = models.CharField(max_length=255, null=True, blank=True)
+    
+    def __str__(self):
+        return self.username
+        
+    def _get_fernet(self):
+        """Get or create a Fernet instance with a key"""
+        if not self._encryption_key:
+            self._encryption_key = Fernet.generate_key().decode()
+            self.save(update_fields=['_encryption_key'])
```
jdamerow: does this mean that the key to decrypt a password is stored in the db along with the password? that does not seem like the best approach. if someone gets access to the database, then they have all they need. might as well not encrypt the password at all. a better approach is probably to store the key as a setting. at least it's separate from the database.

jdamerow (changes requested):

PR #72, opened 2025-03-27
story/VOGRE-67

rajvi-patel-22:

Collections can have subcollections, those also need to be shown
E.g. in dev-test, test1 has two subcollections
[API Documentation | Collections Endpoint](https://diging.atlassian.net/wiki/spaces/CITED/pages/2080899073/API+Documentation#Collections-Endpoint)

https://diging.atlassian.net/browse/VOGRE-67

... describe here ...


annotations/views/repository_views.py
```diff
@@ -313,12 +313,37 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
...
             collection_name = collection_item['name']
+            break
 
     # retrieve items per page from settings and calculate pagination metadata from util function
     items_per_page = settings.PAGINATION_PAGE_SIZE
     pagination = get_pagination_metadata(total_items=texts.get('total_items'), page=page, items_per_page=items_per_page)
 
+    subcollections = []
+    try:
+        subcollections = manager.api.get_group_subcollections(group_id, group_collection_id).get('collections', [])
```
Girik1105: For consistency in the codebase, you can create a manager function that uses the API client to fetch subcollections. example:

```
 def get_subcollections(self, group_id, collection_id):
        """
        Fetch sub‑collections of a given collection in a group.
        """
        return self.api.get_group_subcollections(group_id, collection_id, params=params).get('collections', [])
```
this function would go in the manager class and and then you can directly call the function here.

annotations/views/repository_views.py
```diff
@@ -313,12 +313,37 @@ def repository_collection_texts(request, repository_id, group_id, group_collecti
...
     project_id = request.GET.get('project_id')
+    
+    if not collection_name:
+        parent_collection_key = request.GET.get('parent_collection_key')
+        if parent_collection_key:
+            parent_data = manager.get_subcollections(group_id, parent_collection_key)
+            for collection_item in parent_data:
+                if collection_item['key'] == group_collection_id:
+                    collection_name = collection_item['name']
+                    break
```
jdamerow: this loop can be more pythonic, something like
```
next(collection_item['name'] for collection_item in parent_data if collection_item['key'] == group_collection_id)
```
(haven't tried to run this, but I imagine something like that would work and makes it a lot shorter.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
after that should be fine.

jdamerow:
Make it so, Jenkins.

PR #75, opened 2025-04-03
story/VOGRE-35

rajvi-patel-22:

When clicking on a concept in the concept list if there is not Viaf URL, Vogon should go out to Conceptpower and search first for the concept name, and then for each individual part of the concept name that is longer than 3 characters , and show those results (e.g. Albert Einstein would show results for (int that order):

“Albert Einstein”

“Albert”

“Einstein”

The user can then choose one of those or initiate their own search.
https://diging.atlassian.net/browse/VOGRE-35

... describe here ...


jdamerow:
Resolve conflicts please

annotations/templates/annotations/concept_add.html
```diff
@@ -10,121 +10,245 @@
...
+                                           data-status-target="concept-search-status">
+                                    <span class="input-group-addon">
+                                        <span class="glyphicon glyphicon-search" id="concept-search-status"></span>
+                                    </span>
+                                </div>
+                            </div>
+                            <div class="list-group list-group-compact" id="concept-search-results"></div>
+                        </div>
+                    </div>
+                </div>
```
jophals: I'm not seeing anywhere where you check to see if the concept has a VIAF url

annotations/templates/annotations/concept_add.html
```diff
@@ -10,121 +10,245 @@
...
-        </p>
-    </div>
+    <div class="row" style="display: flex; align-items: flex-start;">
+        <div class="col-md-12">
+            <div class="row" style="display: flex; align-items: flex-start;">
+                <div class="col-md-6" id="search-panel" style="transition: all 0.3s ease;">
+                    <div class="panel panel-default">
+                        <div class="panel-heading">
+                            <h4 class="panel-title">Search for a concept in Conceptpower</h4>
+                        </div>
```
jophals, line 88: I see that you've added the search bar but I'm not seeing the part where Vogon  goes out to ConceptPower to retrieve the results of the concept search and its parts. I'm not sure if you had any conversations with Julia about this but in reading the ticket it looks like the search bar is plan B if the initial results of the ConceptPower retrieval aren't adequate

rajvi-patel-22: So by default, the right panel shows the concepts (and their parts) that we fetch from ConceptPower in the backend ([here](https://github.com/diging/vogon-resuscitate/blob/develop/concepts/views.py#L173)). The left-hand search panel is there if the user wants to look up extra concepts.

Should the search bar only appear when ConceptPower returns no results?

Screenshot 2025-04-24 at 8 46 07 AM

jdamerow:
Make it so, Jenkins.

jdamerow:
Hm, let's rearrange that and put the concept search below the suggested concepts with a heading along the lines "Search Conceptpower for a matching concept."

jdamerow:
Make it so, Jenkins.

PR #76, opened 2025-04-11
story/vogre-7

Girik1105:

Get rid of celery and submit quaduples on request
Description

Currently, quadruples are submitted via Celery on a cron job. Celery should be completely removed. Instead a project should have a button that will submit quadrupes to Quadriga. This ticket includes multiple tasks:
Quadruples need to have a flag (if they don’t already) that indicates if they ahve been submitted or not
Quaduples need to be submitted to the new Quadriga 2.0
Only quadruples with complete relationships (in which concepts have been resolved and added to conceptpower) should be submitted.

[VOGRE-7](https://diging.atlassian.net/browse/VOGRE-7?atlOrigin=eyJpIjoiMDFmYmUyZjU1YzdiNGQ2MDk0ODFmMGE0Y2QyNWE3MDYiLCJwIjoiaiJ9)

... describe here ...

[VOGRE-7]: https://diging.atlassian.net/browse/VOGRE-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annotations/forms.py
```diff
@@ -253,15 +257,95 @@ class RelationTemplateForm(forms.ModelForm):
...
+            node_types = [
+                cleaned_data.get('first_node_type'),
+                cleaned_data.get('second_node_type'),
+                cleaned_data.get('third_node_type')
+            ]
+            
+            node_count = node_types.count('Node')
+            uri_count = node_types.count('URI')
+            
+            if node_count != 2 or uri_count != 1:
```
jdamerow: that's not a requirement. It could be three relation nodes.

annotations/forms.py
```diff
@@ -253,15 +257,95 @@ class RelationTemplateForm(forms.ModelForm):
...
         try:
             [k[1] for k in Formatter().parse(value)]
         except Exception as E:
             raise ValidationError('Invalid expression')
         return value
 
+    def clean(self):
+        cleaned_data = super(RelationTemplateForm, self).clean()
+        
+        if cleaned_data.get('use_relation_nodes'):
```
jdamerow: I'm confused, what is this check for?

annotations/forms.py
```diff
@@ -253,15 +257,95 @@ class RelationTemplateForm(forms.ModelForm):
...
+            # Ensure node values are provided
+            for i, prefix in enumerate(['first', 'second', 'third']):
+                if not cleaned_data.get(f'{prefix}_node_value'):
+                    self.add_error(f'{prefix}_node_value', ValidationError('This field is required'))
+            
+            # Note: We don't build the expression here anymore
+            # It's now handled by JavaScript on form submission
+            
+            # Make sure terminal nodes match with node values
+            if cleaned_data.get('terminal_nodes'):
```
jdamerow: and why is this if beneath the check in line 314?

annotations/forms.py
```diff
@@ -271,16 +355,33 @@ def clean_terminal_nodes(self):
...
-        model = RelationTemplate
-        exclude = ['createdBy']
-
     def __init__(self, *args, **kwargs):
         super(RelationTemplateForm, self).__init__(*args, **kwargs)
         # Set the initial value of 'terminal_nodes' field from the instance's current value for editing form
         if self.instance and hasattr(self.instance, 'terminal_nodes'):
             self.fields['terminal_nodes'].initial = self.instance.terminal_nodes
 
+    def save(self, commit=True):
```
jdamerow, line 402: I'm so confused what this method  is doing.

annotations/models.py
```diff
@@ -1020,6 +1037,17 @@ class RelationTemplate(models.Model):
     expression = models.TextField(null=True)
     """Pattern for representing the relation in normal language."""
 
+    default_mapping = models.TextField(blank=True, null=True)
```
jdamerow: this should be a relationsship to a `DefaultMapping` object that has six properties:
- `subject_type`: relationship node or uri
- `subject_value`: the node reference or a URI
- `predicate_type`: relationship or uri
- ... etc, etc

jdamerow (changes requested):

annotations/forms.py
Girik1105: I have split the form into 2 sections, either the user can enter an expression, or the relation nodes using the dropdown and input fields. This check determines whether to use the standard expression mode or the structured relation nodes mode. When use_relation_nodes is true, the form uses cleans the data from the dropdown fields to fill values such as defaultMapping etc.

annotations/forms.py
Girik1105: This prevents UI-only fields (like use_relation_nodes and the node type/value fields) from being saved to the database model, as these fields exist only for interface functionality used to populate default mapping. Without this filter, Django would try to save these attributes, causing errors since the RelationTemplate model doesn't have these fields. I'll add detailed comments explaining this

jdamerow:
we talked about this during standup

annotations/forms.py
```diff
@@ -253,33 +255,160 @@ class RelationTemplateForm(forms.ModelForm):
...
+        if not first_node_type or not first_node_value:
+            self.add_error('first_node_value', 'First node fields are required')
+            
+        if not second_node_type or not second_node_value:
+            self.add_error('second_node_value', 'Second node fields are required')
+            
+        if not third_node_type or not third_node_value:
+            self.add_error('third_node_value', 'Third node fields are required')
+            
+        return cleaned_data
```
jdamerow, line 377: I believe the above validation can be done by setting a `required` property in the form fields?

annotations/forms.py
```diff
@@ -253,33 +255,160 @@ class RelationTemplateForm(forms.ModelForm):
...
+                 'third_node_type', 'third_node_value']
+    
     def clean_expression(self):
-        from string import Formatter
-        value = self.cleaned_data.get('expression')
-        try:
-            [k[1] for k in Formatter().parse(value)]
-        except Exception as E:
-            raise ValidationError('Invalid expression')
-        return value
```
jdamerow: do we not need these formatters anymore?

annotations/forms.py
```diff
@@ -253,33 +255,160 @@ class RelationTemplateForm(forms.ModelForm):
...
+            self.add_error('second_node_value', 'Second node fields are required')
+            
+        if not third_node_type or not third_node_value:
+            self.add_error('third_node_value', 'Third node fields are required')
+            
+        return cleaned_data
 
     def clean_terminal_nodes(self):
         value = self.cleaned_data.get('terminal_nodes')
+        # Handle empty value for the case where there are no Nodes (all URIs)
```
jdamerow: this field should not be empty

annotations/forms.py
```diff
@@ -253,33 +255,160 @@ class RelationTemplateForm(forms.ModelForm):
...
+            instance.structured_mapping = mapping
+        
+        # Update terminal nodes from relation nodes
+        terminal_nodes = []
+        if first_node_type == 'Node':
+            terminal_nodes.append(first_node_value)
+        if second_node_type == 'Node':
+            terminal_nodes.append(second_node_value)
+        if third_node_type == 'Node':
+            terminal_nodes.append(third_node_value)
```
jdamerow: I don't think this is needed? The terminal nodes should not change (there why are they dependent on the node type?_

annotations/quadriga.py
```diff
@@ -275,3 +266,318 @@ def parse_response(raw_response):
...
+    process_relation(top_relation)
+    
+    #############################################################################
+    # DEFAULT MAPPING SECTION
+    # 
+    # The default mapping is based on the relation template structure.
+    #############################################################################
+    
+    # Initialize an empty default mapping dictionary
+    default_mapping = {}
```
jdamerow: defined further down?

annotations/quadriga.py
```diff
@@ -275,3 +266,318 @@ def parse_response(raw_response):
...
+    if structured_mapping.object_type == DefaultMapping.NODE:
+        default_mapping['object'] = {
+            'type': 'REF',
+            'reference': node_mapping.get(obj_key, "0")
+        }
+    else:  # URI
+        default_mapping['object'] = {
+            'type': 'URI',
+            'uri': structured_mapping.object_value
+        }
```
jdamerow: creating the default mapping should happen in its own function. In general, this function is very long and should be modularized.

annotations/relations.py
```diff
@@ -222,32 +226,96 @@ def parse_template_part_data(part_data, **kwargs):
...
+                terminal_nodes.append(first_node_value)
+            if str(second_node_type) == str(DefaultMapping.NODE):
+                terminal_nodes.append(second_node_value)
+            if str(third_node_type) == str(DefaultMapping.NODE):
+                terminal_nodes.append(third_node_value)
+            
+            if terminal_nodes:
+                template.terminal_nodes = ','.join(terminal_nodes)
+                template.save()
+
```
jdamerow: isn't that code already in forms.py? Why is it here again?

annotations/relations.py
```diff
@@ -503,10 +571,93 @@ def create_relation(template_part, data, relationset, cache={}, appellation_cach
...
+                if '{' + ref + '}' in expression and ref not in calculated_terminal_nodes:
+                    calculated_terminal_nodes.append(ref)
+        
+        # If we have calculated nodes and they differ from provided ones, update
+        if calculated_terminal_nodes:
+            # Check if they're already in the template.terminal_nodes
+            existing_nodes = template.terminal_nodes.split(',') if template.terminal_nodes else []
+            if set(calculated_terminal_nodes) != set(existing_nodes):
+                template.terminal_nodes = ','.join(calculated_terminal_nodes)
+                template.save()
```
jdamerow: similar here and why are terminal nodes related to the default mapping?

jdamerow (changes requested):

annotations/forms.py
Girik1105: I removed this check here since I updated the expression validation to take place on the client side through create.js, the node terminals are auto generated on the client side, meaning when the user inputs expression in the expression field and relation nodes input fields the nodes inputted are automatically filled reducing the room for error.

jdamerow: but terminal nodes are independent of the default mapping (could have different values in there). Also, you can always make a request that sends data to the backend avoiding the front end. In that case there would be no validation.

Girik1105: Okay that makes sense, I'll re-add this validation and re-add the manual entry of terminal nodes in the front end. I'll also remove the validation for default mapping relation nodes and terminal nodes to be same and keep the existing validation between the expression and terminal nodes.

annotations/forms.py
```diff
@@ -225,6 +222,13 @@ def to_python(self, value):
 
 # TODO: widget details (e.g. CSS classes) should be in the template.
 class RelationTemplateForm(forms.ModelForm):
+    # Hidden field to carry the PK of the DefaultMapping for this RelationTemplate (populates structured_mapping)
+    structured_mapping = forms.ModelChoiceField(
+        queryset=DefaultMapping.objects.all(),
```
jdamerow: why does this need to have all default mappings? (especially since it's hidden)

annotations/forms.py
```diff
@@ -263,23 +303,137 @@ def clean_expression(self):
...
+                
+                # Check that all terminal nodes are in expression
+                extra_nodes = []
+                for node in terminal_nodes:
+                    if node not in expression_nodes:
+                        extra_nodes.append(node)
+                
+                if extra_nodes:
+                    raise ValidationError(f"Terminal nodes contain placeholders not found in expression: {', '.join(extra_nodes)}")
+
```
jdamerow: I'm so confused. what is this for? terminal nodes and expression are independent from each other.

jdamerow: terminal nodes would be something like `0s,1o` and expression could be `{0s} is related to {1o} as {0s}`.

annotations/forms.py
```diff
@@ -263,23 +303,137 @@ def clean_expression(self):
...
+                
+                # Check that all placeholders are in node_list
+                missing = [p for p in placeholders if p not in node_list]
+                if missing:
+                    self.add_error('terminal_nodes', f"Terminal nodes must include all placeholders from expression. Missing: {', '.join(missing)}")
+            except Exception as e:
+                # If there was an error in parsing, it should be caught in individual field validations
+                pass
+        
+        return cleaned_data
```
jdamerow, line 377: see above

annotations/models.py
```diff
@@ -1020,6 +1064,11 @@ class RelationTemplate(models.Model):
     expression = models.TextField(null=True)
     """Pattern for representing the relation in normal language."""
 
+    
+    structured_mapping = models.ForeignKey(DefaultMapping, blank=True, null=True, 
+                                          on_delete=models.SET_NULL, related_name='templates')
+    """Structured representation of default mapping relationship configuration."""
```
jdamerow, line 1070: why is this called "structured mapping"? Should be be called "Default Mapping" with some instructions how to fill it in.

annotations/quadriga.py
```diff
@@ -275,3 +265,358 @@ def parse_response(raw_response):
...
+    # Get the subject node from the relation
+    subject_node = None
+    if template_part.source_node_type == RelationTemplatePart.CONCEPT:  # Specific concept
+        subject_node = top_relation.source_content_object
+    elif template_part.source_node_type == RelationTemplatePart.TYPE:  # Open concept
+        subject_node = top_relation.source_content_object
+    elif template_part.source_node_type == RelationTemplatePart.DATE:  # Date
+        subject_node = top_relation.source_content_object
+    elif template_part.source_node_type == RelationTemplatePart.RELATION:  # Relation
+        subject_node = top_relation.source_content_object
```
jdamerow: each if/elif block is exactly the same.

annotations/quadriga.py
```diff
@@ -275,3 +265,358 @@ def parse_response(raw_response):
...
+    # Get the object node from the relation
+    object_node = None
+    if template_part.object_node_type == RelationTemplatePart.CONCEPT:  # Specific concept
+        object_node = top_relation.object_content_object
+    elif template_part.object_node_type == RelationTemplatePart.TYPE:  # Open concept
+        object_node = top_relation.object_content_object
+    elif template_part.object_node_type == RelationTemplatePart.DATE:  # Date
+        object_node = top_relation.object_content_object
+    elif template_part.object_node_type == RelationTemplatePart.RELATION:  # Relation
+        object_node = top_relation.object_content_object
```
jdamerow: same her

annotations/quadriga.py
```diff
@@ -275,3 +265,358 @@ def parse_response(raw_response):
...
+    
+    # Create keys for looking up node IDs
+    subj_key = None
+    obj_key = None
+    
+    if subject_node:
+        if hasattr(subject_node, 'source_content_type_id'):
+            subj_key = f"rel-{subject_node.id}-{subject_node.created.isoformat()}"
+        else:
+            subj_key = f"app-{subject_node.id}-{subject_node.created.isoformat()}"
```
jdamerow: this could be one line:
```
subj_key = f"{'app' if subject_node else 'rel'}-{subject_node.id}-{subject_node.created.isoformat()}"
```
same below

annotations/relations.py
```diff
@@ -157,14 +161,46 @@ def validate_terminal_nodes(template_data, part_data, **kwargs):
...
+            if not node:
+                continue
+            terminal_node_list.append(node)
+            part_id, pred_flag = node
             if not int(part_id) <= N_parts:
                 raise InvalidTemplate("Part ID in terminal nodes is invalid.")
             if not pred_flag in ['s', 'p', 'o']:
                 raise InvalidTemplate("Node ID in terminal nodes is invalid.")
+                
+        # If we have an expression, check that terminal nodes match expression nodes
```
jdamerow: see above

annotations/relations.py
```diff
@@ -222,32 +258,96 @@ def parse_template_part_data(part_data, **kwargs):
...
+            if str(first_node_type) == str(DefaultMapping.NODE):
+                terminal_nodes.append(first_node_value)
+            if str(second_node_type) == str(DefaultMapping.NODE):
+                terminal_nodes.append(second_node_value)
+            if str(third_node_type) == str(DefaultMapping.NODE):
+                terminal_nodes.append(third_node_value)
+            
+            if terminal_nodes:
+                template.terminal_nodes = ','.join(terminal_nodes)
+                template.save()
```
jdamerow: same as above, why are the terminal nodes and default mapping mixed?

annotations/relations.py
```diff
@@ -556,3 +739,57 @@ def update_template(template, template_data, part_data_list):
...
+            # Extract node references from expression
+            formatter = Formatter()
+            refs = [key for _, key, _, _ in formatter.parse(expression) if key is not None]
+            
+            # Ensure refs from expression match terminal nodes
+            for ref in refs:
+                if '{' + ref + '}' in expression and ref not in parsed_nodes:
+                    raise ValidationError(f"Terminal nodes must include all nodes referenced in the expression. Missing: {ref}")
+        
+        return value
```
jdamerow: same as before. terminal nodes and default mapping are very different. they are independent of each other.

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

annotations/forms.py
```diff
@@ -264,22 +308,145 @@ def clean_expression(self):
...
+                for _, field, _, _ in Formatter().parse(expression):
+                    if field and not re.match(r'^\d+[spo]$', field):
+                        self.add_error('expression', f"Invalid placeholder format: {{{field}}}. Should be a number followed by 's', 'p', or 'o'.")
+            except Exception as e:
+                self.add_error('expression', "Invalid expression format")
+        
+        # Validate Node/URI field pairs
+        first_node_type = cleaned_data.get('first_node_type')
+        first_node_value = cleaned_data.get('first_node_value')
+        print(first_node_type, first_node_value)
```
jdamerow: if these are supposed to be log messages, they need more info, otherwise remove the print statements.

annotations/models.py
jdamerow: this still

annotations/relations.py
```diff
@@ -152,46 +156,102 @@ def build_dependency_graph(template_data, part_data, **kwargs):
...
+                    # verify the key has an index and a field identifier
+                    try:
+                        part_id = int(key[0])
+                        field_id = key[-1]
+                        if field_id not in ['s', 'p', 'o']:
+                            raise ValueError()
+                    except (ValueError, IndexError):
+                        raise InvalidTemplate(f"Invalid key format in expression: {{{key}}}")
+        except Exception as E:
+            raise InvalidTemplate("Invalid expression format")
```
jdamerow: is there a reason that the above functions are no longer used? They are still there right?

annotations/relations.py
```diff
@@ -152,46 +156,102 @@ def build_dependency_graph(template_data, part_data, **kwargs):
...
+    # Validate terminal nodes
+    if 'terminal_nodes' in template_data and template_data['terminal_nodes']:
+        try:
+            node_list = template_data['terminal_nodes'].split(',')
+            for node in node_list:
+                node = node.strip()
+                if not node:
+                    continue
+                if len(node) < 2 or not node[0].isdigit() or node[-1] not in ['s', 'p', 'o']:
+                    raise InvalidTemplate(f"Invalid node format in terminal_nodes: {node}")
```
jdamerow: I know this is the content of the original validation function (which should not copied here but called, see comment below), but this doesn't seem completely right.
1: what if someone enters `,1o`? that would be an incorrect pattern but it would be caught, would it?
2: what if someone enters `1oo`, this would be an incorrect pattern that would not be caught either, would it?

annotations/relations.py
jdamerow: wait, isn't that pretty much the same code as validating the nodes?

annotations/utils.py
```diff
@@ -8,6 +8,26 @@
...
+    delimiter : unicode
+        Character or sequence by which to split and join tokens.
+
+    Returns
+    -------
+    tokenizedContent : unicode
+    """
+    chunks = content.split(delimiter) if content is not None else []
+    pattern = '<word id="{0}">{1}</word>'
+    return delimiter.join([pattern.format(i, c) for i, c in enumerate(chunks)])
```
jdamerow, line 30: is this part of this ticket?

jdamerow (changes requested):

annotations/utils.py
Girik1105: So this PR is also includes [VOGRE-29](https://diging.atlassian.net/browse/VOGRE-29?atlOrigin=eyJpIjoiNDNlNmI2Yjc0ZmQzNDhiNThjYTQyMDRlOTIzMjdjODkiLCJwIjoiaiJ9). The tokenize function was first in the tasks.py file so I shifted it to utils

annotations/relations.py
Girik1105: Im not sure which function you're referring to, a lot of the functions above are used in the update template function which is used in relation template_views

jdamerow: line 191-192 deleted:
```
validate_terminal_nodes(template_data, part_data)
validate_expression(template_data, part_data)
```

Girik1105: These functions are not being used, I just pushed another commit removing them, the equivalent logic is being executed in validate_template_data function.

annotations/relations.py
jdamerow: can't you use the same regex expression you used earlier to validate nodes?

annotations/relations.py
jdamerow: ja, but why? why make it less modular?

vogon/settings.py
```diff
@@ -260,9 +247,20 @@
...
 BASE_URL = os.path.join(os.getenv('BASE_URL', '/'), APP_ROOT)
 
+# Conceptpower Credentials
+CONCEPTPOWER_USERID = os.environ.get('CONCEPTPOWER_USERID')
+CONCEPTPOWER_PASSWORD = os.environ.get('CONCEPTPOWER_PASSWORD')
+CONCEPTPOWER_ENDPOINT = os.environ.get('CONCEPTPOWER_ENDPOINT')
+CONCEPTPOWER_NAMESPACE = os.environ.get('CONCEPTPOWER_NAMESPACE', 'http://www.digitalhps.org/')
+
+QUADRIGA_ENDPOINT = os.environ.get('QUADRIGA_ENDPOINT')
+QUADRIGA_CREATION_PLACE = "Phoenix, Az"
```
jdamerow: AZ not Az

jdamerow (changes requested):

annotations/relations.py
Girik1105: The new approach is more restrictive and therefore a stronger safeguard because It enforces exactly 2-character keys (prevents multi-digits and it also prevents users from accidentally entering two commas  an issue you referred to before) E.x. 1o,,2o.

The regex I used `/^(\d)([spo])$` would let multiple digits as one node therefore I changed the validation.

jdamerow:
Make it so, Jenkins.

jdamerow:
when trying to save a default mapping to an existing template, I get this error:
image
image

PR #77, opened 2025-04-17
story/VOGRE-73

Girik1105:

Vogon should be able to annotate TEI-XML files

[TEI](https://www.tei-c.org/) is an XML format to annotate texts (it defines what paragraphs there are, line breaks, etc, see example). Vogon should be able to show the text (without the XML markup) with proper layout (respecting paragraphs, line breaks, etc), and then let the user annotate the text. Here, pointers would need to be xpointers (referencing by xpath and character count within tag probably).

[VOGRE-73](https://diging.atlassian.net/browse/VOGRE-73?atlOrigin=eyJpIjoiYWI1N2E3ZjE1NGFjNGY5YmI5NTk1ZTJkOTE4MjE0ZjIiLCJwIjoiaiJ9)

... describe here ...

[VOGRE-73]: https://diging.atlassian.net/browse/VOGRE-73?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jophals:
Am i doing something wrong? I used the example xml file in Jira but my text in the annotation view still had all the xml tags and headers 
Screenshot 2025-04-21 at 11 12 05 PM

Girik1105:
> Am i doing something wrong? I used the example xml file in Jira but my text in the annotation view still had all the xml tags and headers Screenshot 2025-04-21 at 11 12 05 PM

I was missing checks for some namespace tags, I have added code in our TEI-XML utility files to strip the namespace tags and display the text in appropriate formatting

annotations/annotators.py
```diff
@@ -196,6 +197,95 @@ def get_context(self):
...
+            return None
+            
+        content_info = tei_utils.detect_content_type(resource)
+        
+        if content_info['is_tei']:
+            tei_data = tei_utils.parse_tei_document(resource)
+            
+            tokenized_content = tei_utils.tokenize_tei_content(tei_data['display_html'])
+
+            self.tei_data = tei_data
```
jdamerow: merge with line 241

annotations/tei_utils.py
```diff
@@ -0,0 +1,585 @@
...
+
+    Returns a dictionary with:
+        - is_xml (bool)
+        - is_tei (bool)
+        - content_type (str) = 'text/xml+tei', 'text/xml', or 'text/plain'
+    """
+    is_xml = False
+    is_tei = False
+
+    if text_content.strip().startswith('<?xml') or text_content.strip().startswith('<'):
```
jdamerow: The first condition is included in the second, isn't it? If the text starts with "<?xml", then it also starts with "<".

annotations/tei_utils.py
```diff
@@ -0,0 +1,585 @@
...
+      NOT children (.text / .tail) are skipped – they don't belong
+      in the cleaned tree and would otherwise raise TypeError
+      when appended.
+    """
+    # Skip non-element nodes (like processing instructions) to prevent errors
+    if not isinstance(node, etree._Element):
+        return None
+
+    # ── create namespace-free copy of *this* element ────────────────────
+    if node.tag.startswith('{'):
```
jdamerow: when would a tag name start with `{`?

annotations/tei_utils.py
```diff
@@ -0,0 +1,585 @@
...
+        - clean_xml: namespace-stripped XML
+        - display_html: improved HTML representation
+        - element_map: mapping between display positions and TEI elements
+        - tei_metadata: extracted metadata from TEI header
+        - css: placeholder for CSS (if any)
+        - facsimile_data: info about facsimile images
+    """
+    try:
+        # Use a parser that removes processing instructions (XML declarations, etc.)
+        # This prevents issues with <?xml ...?> and <?xml-model ...?> tags
```
jdamerow: why are there issues with those tags?

annotations/tei_utils.py
jdamerow: ah I see, a better way is probably to do what the documentation suggests and to include namespaces when searching the xml: https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml-with-namespaces

annotations/tei_utils.py
```diff
@@ -0,0 +1,585 @@
...
+
+    # Handle body and text containers
+    if tg in ('body', 'text'):
+        # Create a containing element but process children
+        html_parts.append(f'<div class="tei-{tg}" data-xpath="{xpth}">')
+        element_map.append({'xpath': xpth, 'element_type': tg,
+                            'start_pos': len("".join(html_parts))})
+        if element.text:
+            html_parts.append(_escape_html(element.text))
+        for ch in element:
```
jdamerow: what does `ch` stand for?

annotations/tei_utils.py
```diff
@@ -0,0 +1,585 @@
...
+                        'start_pos': len("".join(html_parts))})
+    if element.text:
+        html_parts.append(_escape_html(element.text))
+    for ch in element:
+        process_element(ch, html_parts, element_map, xpth)
+        if ch.tail:
+            html_parts.append(_escape_html(ch.tail))
+    html_parts.append(f'</span>')
+    
+    return {'html_parts': html_parts, 'element_map': element_map}
```
jdamerow: this needs to be taken apart. The most elegant solutions is probably to create a method for each tag, something like `_process_cell` or `_process_item` and then dynamically call the methods given the tag.

jdamerow (changes requested):

annotations/tei_utils.py
Girik1105: image

Before this try/catch block, XML processing instructions like `<?xml ...?>` slipped through the parser. The parser is now configured with remove_pis=True (remove processing instructions), which specifically removes XML declarations and other processing instructions like` <?xml ...?> `and` <?xml-model ...?>. `

annotations/tei_utils.py
Girik1105: This was short for child_elements, I have changed it to use better variable names

annotations/tei_utils.py
jdamerow: I don't think I understand the comment then. Are you trying to tell people, they should be using a parser that removes those tags, but if not, this try/catch will take care of it?

jdamerow: wait, not, the parser comes after. I'm confused.

annotations/tei_utils.py
jdamerow: this still

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

annotations/tei_utils.py
```diff
@@ -0,0 +1,1138 @@
...
+    
+    # Extract namespace map from root element
+    namespace_map = root.nsmap.copy() if root.nsmap else {}
+    
+    # Find the TEI namespace
+    tei_namespace_uri = None
+    for prefix, uri in namespace_map.items():
+        if 'tei-c.org' in uri:
+            tei_namespace_uri = uri
+            break
```
jdamerow, line 89: if you want to properly parse the xml, you'll need all namespaces, not just the TEI one I would think

annotations/tei_utils.py
```diff
@@ -0,0 +1,1138 @@
...
+    
+    # If root is TEI but no namespace declared, assume default
+    if not tei_namespace_uri and etree.QName(root).localname in ['TEI', 'tei']:
+        tei_namespace_uri = TEI_NAMESPACE
+        namespace_map[None] = TEI_NAMESPACE
```
jdamerow, line 98: this should not be possible. If a prefix is used, then the prefix needs to be declared with the namespace uri i believe. Or is this doing something else?

jdamerow: and if there is a default namespace (elements are used without a prefix), then the namespace is specified with the `xmlns` attribute i believe.

jdamerow (changes requested):

jdamerow:
also, there are conflicts

PR #79, opened 2025-04-22
story/VOGRE-72

rajvi-patel-22:

So, e.g.:

a user flags a concept

another user (or the same user) adds a response to the comment (this could happen multiple times)

comments should be ordered chronologically, so you can follow the conversation
https://diging.atlassian.net/browse/VOGRE-72

... describe here ...


concepts/views.py
```diff
@@ -210,3 +210,74 @@ def sandbox(request, text_id):
...
+@login_required
+def delete_comment(request, comment_id):
+    comment = get_object_or_404(Comment, pk=comment_id)
+    
+    if request.user != comment.created_by:
+        messages.error(request, 'You do not have permission to delete this comment.')
+        return redirect('concept', concept_id=comment.concept.id)
+
+    if request.method == 'POST':
+        concept_id = comment.concept.id
```
jdamerow, line 281: merge with line 281

concepts/views.py
```diff
@@ -210,3 +210,74 @@ def sandbox(request, text_id):
     from annotations.models import RelationTemplate
 
     return render(request, "annotations/relationtemplate_creator.html", {})
+
+@login_required
+def add_comment_reply(request, comment_id):
```
jdamerow, line 217: pretty much all these methods should check if the user has access to the concept. Just because a concept is not shown, doesn't mean one could construct a request that adds a comment to a concept (e.g. guessing the concept or comment id).

jdamerow (changes requested):

concepts/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+        comment_id = kwargs.get('comment_id')
+        
+        if concept_id:
+            concept = get_object_or_404(Concept, pk=concept_id)
+        elif comment_id:
+            comment = get_object_or_404(Comment, pk=comment_id)
+            concept = comment.concept
+            
+        if concept and (request.user == concept.createdBy or 
+                       request.user.is_vogon_admin or request.user.is_staff):
```
jdamerow, line 27: I don't think we use `is_staff`, do we? Shouldn't it be `is_superuser`?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

concepts/decorators.py
rajvi-patel-22: Most of the places I've checked use `is_staff`. I haven’t come across any usage of `is_superuser`.

jdamerow: just leave it `is_staff` for now

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

